PR #11, opened 2016-01-30
Force django to use settings.LANGUAGE_CODE

saeedmb8:


gem/settings/base.py
```diff
@@ -122,7 +123,7 @@
 # Internationalization
 # https://docs.djangoproject.com/en/1.7/topics/i18n/
 
-LANGUAGE_CODE = 'en-gb'
+LANGUAGE_CODE = 'id'
```
miltontony: Please leave this as is. It will get set on the server.

miltontony:
:+1:

PR #23, opened 2016-03-22
Header markup and style tweaks

peterp:
1. Removed all block element in favour of inline elements. (Removed left and right floating.)
2. Used line-height to suggest height.
3. Used `` instead of css background images.
4. Vertically aligned logo and text.

screen shot 2016-03-22 at 23 33 43


miltontony:
:+1:

PR #24, opened 2016-03-23
New Indonesian translations

peterp:
I had to translate some molo template strings, I just extracted them from site-packages.


gem/templates/profiles/viewprofile.html
```diff
@@ -4,7 +4,7 @@
 {% block content %}
     <div class="profile-container">
         <div class="title profile">
-            <h3>{% trans "Hello " %}{{ user }}</h3>
+            <h3>{% trans "Hello " %} {{ user }}</h3>
```
smn: The space already seems to be in the translation string :confused:

miltontony: yeah, lets remove the space from the translation. It should just be `{% trans "Hello" %}`

locale/id/LC_MESSAGES/django.po
```diff
@@ -181,94 +175,96 @@ msgid "Edit your Profile"
...
 
-#: gem/templates/profiles/register.html:26
+#: gem/templates/profiles/register.html:27
 msgid "already joined?"
 msgstr "Sudah bergabung?"
 
-#: gem/templates/profiles/viewprofile.html:6
-msgid "Hello "
+#: gem/templates/profiles/viewprofile.html:7
+#, fuzzy
```
miltontony: Please delete this line - basically django is confirming that it did a partial match and found these 2 strings to still be the same. If you leave this in, `+#, fuzzy` it ignores this when the translation compiles.

miltontony:
:+1:

PR #26, opened 2016-03-23
Made nav bar match designs.

peterp:
Original:
screen shot 2016-03-23 at 16 33 48

New Feature:

screen shot 2016-03-23 at 16 34 18

New Smart: 
screen shot 2016-03-23 at 16 34 13


saeedmb8:
:+1:

PR #27, opened 2016-03-23
removed google web fonts.

peterp:
1. Removed google web fonts
2. Removed JavaScript that was loading main.js, which was empty.
3. Removed extra_js, and extra_css blocks. (We don't use them.)


miltontony:
:+1:

PR #31, opened 2016-03-29
Button styles, section styles

peterp:
- Removed individual styles for buttons and centralised them into a single block.
- Sharing code that generates `section headers`
- Some template tag cleanups: removing extra span tags.


smn:
@miltontony or @saeedmb8 could either of you review?

PR #32, opened 2016-03-29
Sharing article preview code between homepage, section list and article.

peterp:
Sharing code and class names across different sections of the website. We had 3 views with almost identical code, but none of it was identical enough to reason about.


smn:
:cake: :+1:

saeedmb8:
:+1:

PR #36, opened 2016-03-30
Added 16:9 and square images, updated quality

peterp:


gem/static/css/enhanced.scss
```diff
@@ -47,7 +47,7 @@
 body {
   font-size: 18px;
   line-height: 24px;
-  max-width: 320px;
+  max-width: 400px;
 }
```
saeedmb8: @peterp Is there any specific reason that we are changing this to `400px`?

peterp: @saeedmb8 for phones in landscape; ideally this should actually just resize to 100%, and take up all the available space on the device. But we're busy showing the site to people and they get confused when it takes up all the space.

peterp: @saeedmb8 what do you guys typically do?

saeedmb8:
:+1:

PR #47, opened 2016-04-05
GEM-138 Updated pattern image

Pringels:
@saeedmb8 @KaitCrawford 

Trimmed 2 empty pixels from pattern image and reduced container height to fit.
Pattern now sits flush with footer box.


saeedmb8:
:+1:

PR #49, opened 2016-04-05
GEM-136 Set divider images to auto scale

Pringels:
@saeedmb8 @KaitCrawford 

The divider images need to align perfectly across all viewport widths.
Not the most elegant solution but it does the trick.


saeedmb8:
:+1:

PR #50, opened 2016-04-06
GEM-139 Added styling to article category links in footer menu

Pringels:
@saeedmb8 @KaitCrawford 

This is the solution Indrani and I decided on instead of using a dropdown.


saeedmb8:
:+1:

PR #51, opened 2016-04-06
GEM-150 Changed article section header color to grey. Reduced font size.

Pringels:
@saeedmb8 @KaitCrawford 

Also removed section icon on article page. (bug introduced in GEM-137)


saeedmb8:
:+1:

PR #56, opened 2016-04-07
GEM-154 Fixed cropping of images

Pringels:
@saeedmb8 

I changed the image template tag to always pull in square images.


saeedmb8:
:+1:

PR #62, opened 2016-04-14
Manually added translations

peterp:
This is not a long term solution. We need a way to include translations strings from molo.* and django.contrib.auth, etc.


saeedmb8:
👍

PR #65, opened 2016-04-15
Add gender field on registration

KaitCrawford:


gem/templates/profiles/viewprofile.html
```diff
@@ -65,5 +65,11 @@
                 </form>
             </div>
         </div>
+        <hr>
+        <div class="gender-form">
+            <div class="block">
+                    <h5>{% trans "Your gender is:" %}{{user.gem_profile.gender}}</h5>
```
KaitCrawford: Currently this isn't working and I'm not sure why. `user` prints the username and `user.profile` prints the profile object but `user.gem_profile` doesn't print anything.

KaitCrawford: Nevermind. I was logged in with a user that didn't have a gender specified.

gem/forms.py
```diff
@@ -0,0 +1,50 @@
...
+        widget=forms.TextInput(
+            attrs=dict(
+                required=True,
+                max_length=30,
+            )
+        ),
+        label=_("Username"),
+        error_messages={
+            'invalid': _("This value must contain only letters, "
+                         "numbers and underscores."),
```
rudigiesler: According to the regex above, also `.`, `@`, `+`, and `-`.

gem/templates/profiles/register.html
```diff
@@ -19,6 +19,10 @@
             <p><strong>{% trans "Choose a 4-digit pin" %}</strong></p>
             {{ form.password.errors}}
             {{ form.password }}
+
+            <p><strong>{% trans "What gender are you" %}</strong></p>
```
rudigiesler: We probably want `Choose your gender` to keep it the same as the other questions.

gem/templates/profiles/viewprofile.html
```diff
@@ -65,5 +65,13 @@
                 </form>
             </div>
         </div>
+        {% if user.gem_profile %}
+          <hr>
+          <div class="gender-form">
+              <div class="block">
+                      <h5>{% trans "Your gender is:" %}{{user.gem_profile.gender}}</h5>
```
rudigiesler: We probably want `male` here instead of `m`.

gem/constants.py
```diff
@@ -0,0 +1,3 @@
+MALE = "m"
+FEMALE = "f"
+GENDERS = {(MALE, "male"), (FEMALE, "female")}
```
rudigiesler: At minimum, we probably want at least `other` and `prefer not to say` choices as well.

gem/tests/test_forms.py
```diff
@@ -0,0 +1,55 @@
...
+            'username': 'Jeyabal#',
+            'password': '1234',
+            'gender': 'f',
+        }
+        form = RegistrationForm(data=form_data)
+        self.assertEqual(form.is_valid(), False)
+
+    def test_register_password_incorrect(self):
+        form_data = {
+            'username': 'Jeyabal#',
```
rudigiesler: By using a bad username here, the form will be incorrect, even if the password is correct.

KaitCrawford:
@miltontony @hodgestar @saeedmb8 ready for review

gem/forms.py
KaitCrawford: hmm. What's valid regex for preventing those?

gem/templates/profiles/viewprofile.html
KaitCrawford: @rudigiesler I'm just not sure how to get the constant variables into the context here

gem/forms.py
rudigiesler: For just letters, numbers, and underscores, you just need  `^\w+$`

gem/templates/profiles/viewprofile.html
rudigiesler: You probably need a template tag? That converts the `m` into `male`, etc. Then you can `{{user.gem_profile.gender|expand_gender}}` or something.

gem/forms.py
KaitCrawford: Leaving this for now as it isn't relevant to the goal of this pull request and needs some discussion (since this is the stolen from molo.profiles it's a decision that will need to be changed there as well)

gem/constants.py
KaitCrawford: After discussion on Slack I've added Trans as an option and indicated when registering that your gender will only be visible to you.

KaitCrawford:
@rudigiesler @miltontony @hodgestar @saeedmb8 ready for re-review

KaitCrawford:
@rudigiesler @miltontony @hodgestar @saeedmb8 ready for re-review

gem/forms.py
rudigiesler: I think this regex was build for email addresses? Either way, if the user follows the error message they'll get a valid username, so I'm okay with leaving it for now.

gem/forms.py
```diff
@@ -0,0 +1,50 @@
+from django import forms
+from django.contrib.auth.models import User
+from django.utils.translation import ugettext_lazy as _
+from gem.constants import GENDERS
+
+
+class RegistrationForm(forms.Form):
```
miltontony: @KaitCrawford I think you can just inherit from `molo.profiles.forms` - I noted it for TuneMe after the fact. In which case you only need to specify the Gender field.

KaitCrawford: @miltontony done in the latest commit

miltontony:
:+1: happy!

PR #72, opened 2016-04-22
Update web fonts + Translations

peterp:
1. Fixed an error in a translation.
2. Updated the web-fonts to use the files that are licensed.


locale/id/LC_MESSAGES/django.po
```diff
@@ -24,7 +24,7 @@ msgstr "perempuan"
 
 #: gem/constants.py:6
 msgid "don't want to answer"
```
miltontony: Looks like Pootle is behind, we'll need to update it.

miltontony:
:+1:

locale/id/LC_MESSAGES/django.po
peterp: @miltontony maybe a great goal for automation? E.g.: add some functionality to pootle to update from `develop`

PR #73, opened 2016-04-25
Feature/gem 194 display birthday in profile page

RichardLanders:


miltontony:
@RichardLanders this PR is invalid. The date of birth field already exists in `molo.profiles`
This PR needs to be reverted. Unfortunately since this includes a migration, you can't simply delete migration 0002, you'll have to generate a new migration for 003.

PR #81, opened 2016-04-28
Feature/gem 192 rss feed

maltem-za:


gem/views.py
```diff
@@ -60,3 +62,28 @@ def form_valid(self, form):
...
+
+class GemRssFeed(Feed):
+    title = 'GEM Feed'
+    link = '/feed/'
+    description = 'GEM Feed'
+    description_template = 'feed_description.html'
+
+    def items(self):
+        return ArticlePage.objects.filter(
+            live=True
```
miltontony: Here you can use `ArticlePage.objects.live()`

miltontony:
Looks good, but can you add a simple test for both the atom and the rss feeds. Just need to make sure that those URLs are covered.

gem/views.py
maltem-za: Thanks, done

maltem-za:
@miltontony Please review :)

gem/views.py
```diff
@@ -60,3 +62,53 @@ def form_valid(self, form):
...
+    description_template = 'feed_description.html'
+
+    def __call__(self, request, *args, **kwargs):
+        self.base_url = '{0}://{1}'.format(request.scheme, request.get_host())
+        return super(GemRssFeed, self).__call__(request, *args, **kwargs)
+
+    def get_feed(self, obj, request):
+        feed = super(GemRssFeed, self).get_feed(obj, request)
+        # set language to Tagalog instead of settings.LANGUAGE_CODE
+        feed.feed['language'] = 'tl'
```
miltontony: Hmm..this is not going to work - we have 5 sites that are based on this repo.
I think you can use 
```python
language = SiteLanguage.objects.filter(is_main_language=True).first()
if language:
    feed.feed['language'] = language.locale
else:
    feed.feed['language'] = 'en'
```

miltontony: :( sorry, this is only available in the new stuff that's unreleased yet.

For now you'll have to leave it as `settings.LANGUAGE_CODE`
We set this for all projects so it should work as expected.

maltem-za: Ok, done as requested

miltontony:
:+1:

PR #86, opened 2016-05-04
Rwanda Style Problems

peterp:
- fixed some issues with the patterns.

screen shot 2016-05-04 at 12 22 19
screen shot 2016-05-04 at 12 22 26


saeedmb8:
👍

PR #88, opened 2016-05-06
Feature/gem 201 reject comments that contain mobile numbers or email address

RichardLanders:


RichardLanders:
@miltontony please review

gem/models.py
```diff
@@ -18,3 +21,21 @@ def gem_user_profile_handler(sender, instance, created, **kwargs):
     if created:
         profile = GemUserProfile(user=instance)
         profile.save()
+
+
+@register_setting
+class SiteSettings(BaseSetting):
```
miltontony: Let's call this "GemSettings" as SiteSettings already exists in `molo.core`

gem/models.py
```diff
@@ -18,3 +21,21 @@ def gem_user_profile_handler(sender, instance, created, **kwargs):
...
+    comment_filters = models.ForeignKey(
```
miltontony: This doesn't appear to be used anywhere

gem/views.py
```diff
@@ -114,3 +121,35 @@ def item_author_name(self, article_page):
...
+        settings.banned_keywords_and_patterns.split('\n') \
+        if settings.banned_keywords_and_patterns else []
+
+    banned_list += banned_keywords_and_patterns
+
+    for keyword in banned_list:
+        keyword = keyword.replace('\r', '')
+        match = re.search(keyword.lower(), comment.lower())
+        if match:
+            raise forms.ValidationError("Please remove the following "
```
miltontony: This error needs to be wrapped in `_(..)` for translations.
We'll need to check with Indrani on whether we want

gem/models.py
maltem-za: Done

gem/models.py
maltem-za: Removed

gem/views.py
maltem-za: Done

gem/views.py
```diff
@@ -304,3 +312,38 @@ def item_author_name(self, article_page):
...
+                _(
+                    'This comment has been removed as it contains profanity, '
+                    'contact information or other inappropriate content. '
+                )
+            )
+
+    return comment
+
+
+CommentDetailsForm.clean_comment = clean_comment
```
smn: Wouldn't it be simpler to subclass the comment form? This kind of patching feels like asking for trouble.

smn:
:+1: in general from me, just worried about the monkey patching going on in the CommentDetailsForm. What other options have been considered before settling on this?

gem/views.py
```diff
@@ -145,9 +146,12 @@ def clean_comment(self):
         keyword = keyword.replace('\r', '')
         match = re.search(keyword.lower(), comment.lower())
```
maltem-za: @smn @miltontony How do you guys feel about `keyword.lower()`? I'm a bit concerned as this will change the meaning of regexes with special sequences (eg: \D). I'd be more comfortable not lowering the keyword and instructing users to write keywords as lowercase in the help text for the field on the GemSettings model.

gem/views.py
```diff
@@ -304,3 +312,38 @@ def item_author_name(self, article_page):
...
+
+    banned_keywords_and_patterns = \
+        settings.banned_keywords_and_patterns.split('\n') \
+        if settings.banned_keywords_and_patterns else []
+
+    banned_list += banned_keywords_and_patterns
+
+    for keyword in banned_list:
+        keyword = keyword.replace('\r', '')
+        match = re.search(keyword.lower(), comment.lower())
```
maltem-za: @smn @miltontony How do you guys feel about keyword.lower()? I'm a bit concerned as this will change the meaning of regexes with special sequences (eg: \D). I'd be more comfortable not lowering the keyword and instructing users to write keywords as lowercase in the help text for the field on the GemSettings model.

gem/views.py
maltem-za: @smn I've DM'ed you on Slack about this

gem/views.py
miltontony: I'm happy to remove `.lower` and rely on regex. It is _more correct_ that way.

gem/views.py
miltontony: I agree, this patching is very dirty and the reference I gave (from YAL) had a similar note.
I'd also recommend a cleaner approach, but I haven't applied my mind yet as to what that _cleaner approach_ would be.

gem/views.py
maltem-za: Done

PR #95, opened 2016-05-30
Update Settings and Docker files to Molo 3 and Django 1.9

saeedmb8:


.deploy.yaml
```diff
@@ -1,5 +1,5 @@
-buildscript: sideloader/build.sh
-postinstall: sideloader/postinstall.sh
+buildscript: gem/sideloader/build.sh
+postinstall: gem/sideloader/postinstall.sh
```
miltontony: This should not change

docker/docker-start.sh
```diff
@@ -5,7 +5,8 @@ set -e
 # ensure we create the locales first so that Django doesn't complain
 LANGUAGE_CODE=en ./manage.py compilemessages
 
-./manage.py migrate
+./manage.py migrate core 0031 --fake
```
miltontony: Please add a note as to why this is done

docker/settings.py
```diff
@@ -4,7 +4,7 @@
 
 import dj_database_url
 
-from .base import *
+from .production import *
```
miltontony: This should remain as `.base`

miltontony:
Don't delete the TAG Manager settings value - use wagtail settings as the first prize and fallback to the settings value. We can remove it at a later stage - create an issue for the removal.

saeedmb8:
@miltontony ready for review, I will Update the rest of the templates in a separate PR

gem/context_processors.py
```diff
@@ -12,6 +13,8 @@ def default_forms(request):
 
 
 def add_tag_manager_account(request):
+    GTM = SiteSettings.for_site(request.site)
```
miltontony: this variable should be `site_settings` instead of `GTM`

gem/context_processors.py
```diff
@@ -12,6 +13,8 @@ def default_forms(request):
...
     return {
-        'GOOGLE_TAG_MANAGER_ACCOUNT': settings.GOOGLE_TAG_MANAGER_ACCOUNT
+        'GOOGLE_TAG_MANAGER_ACCOUNT': (settings.GOOGLE_TAG_MANAGER_ACCOUNT or
```
miltontony: Please swap this 2 so that the wagtail value is  tried first

miltontony:
👍

PR #99, opened 2016-06-08
make sure languages show the correct conent

saeedmb8:


gem/settings/base.py
```diff
@@ -288,9 +291,9 @@
 
 # Password reset - security questions
 SECURITY_QUESTION_1 = environ.get(
-    'SECURITY_QUESTION_1', 'Security question 1?')
+    'SECURITY_QUESTION_1', 'Name ofthe city you were born in')
```
hodgestar: s/ofthe/of the/ ?

saeedmb8: Thanks @hodgestar I also noticed it after I deployed, will fix it in the morning once I got the translations :)

PR #101, opened 2016-06-13
Improved UX of registration

peterp:
@Mitso Can I get a +1?

Better error message, label is tappable, and looks like a link for checkbox.
screen shot 2016-06-13 at 15 18 16

Example of a PIN:
screen shot 2016-06-13 at 15 13 57

### DIFF
--- a/gem/static/css/style.scss
+++ b/gem/static/css/style.scss
@@ -212,6 +212,20 @@ a, a:hover {
   }
 }
 
+
+.signin-form input[type="checkbox"] {
+  padding: 5px;
+}
+
+.terms-and-conditions label {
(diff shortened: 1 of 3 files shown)


Mitso:
@peterp the PIN example could be next to the label or part of the label for better UX as well. Both options work though just something to consider if not already :) 
+1

peterp:
@Mitso like this?

screen shot 2016-06-13 at 15 35 58

Mitso:
Actually worked well. Meant something like this:
screenshot

peterp:
Aaaah, great call!

Mitso:
Great! +1

PR #102, opened 2016-06-14
Make Current Language Visible - GEM MOLO (FED)

Mitso:


Mitso:
@peterp please review

peterp:
👍

PR #104, opened 2016-06-14
Pattern in the last section missing accent V Mark - GEM MOLO (FED)

Mitso:


Mitso:
@peterp please review 

Old Pattern
pattern_bar_old

New Pattern
pattern_bar_old

peterp:
@Mitso I think we need to fix the image on this one, if you look at the last part of the pattern... it doesn't have the "V" mark in it. Let me know if you need help.

Mitso:
@peterp please see the image example
http://gem.molo.unicore.io/static/img/featured_bar.png?268d49d83da5

My thinking is, the image pattern starts and ends without v, then it has repeat-x: 
Solution is the above or we can expand it like this:
pattern_new_

peterp:
@Mitso 

I've modified the pattern to be repeatable

pattern

Mitso:
@peterp oh great! thanks :D

peterp:
👍

PR #108, opened 2016-06-21
fixed user genders not displaying on django-admin

RichardLanders:
Minor tweak to get genders displaying again on the django-admin


saeedmb8:
👍

PR #119, opened 2016-06-30
Feature/gem 292 report reason form

RichardLanders:


RichardLanders:
@saeedmb8 @miltontony

gem/views.py
```diff
@@ -377,3 +377,60 @@ def clean_comment(self):
...
+    def render_to_response(self, context, **response_kwargs):
+        comment = MoloComment.objects.get(pk=self.kwargs['comment_pk'])
+
+        for report in comment.gem_comment.all():
+            if self.request.user.id == report.user_id:
+                return HttpResponseRedirect(
+                    reverse('already_reported',
+                            args=(self.kwargs['comment_pk'],)
+                            )
+                )
```
maltem-za: 1. The property name `gem_comment` on `MoloComment` is a bit confusing. It sounds like it's referring to another type of comment but actually it's referencing the `GemReportComment` model. I would change the `related_name` on the `comment` property of `GemReportComment` to something like `gem_comment_reports` (I think the default, if you didn't set the `related_name`, would be `gemreportcomment_set`? - which would also be fine). Read on before you do this.

2. Consider changing the model name from `GemReportComment` to `GemCommentReport`. It reads better and will probably take care of the above if you let it use the default `related_name`. `gemcommentreport_set` is pretty descriptive when you're looking at a `MoloComment` instance. If you do this rename please also rename the related `GemReportCommentModelAdmin` and `GemReportCommentAdmin` classes for consistency and remember to update your migration.

3. You should be able to filter for the user instead of looping over every report and checking the user id.

gem/views.py
```diff
@@ -377,3 +377,60 @@ def clean_comment(self):
...
+                reverse(
+                    'molo-comments-report', args=(self.kwargs['comment_pk'],)
+                ),
+                reverse(
+                    'report_response', args=(self.kwargs['comment_pk'],))
+            )
+        )
+
+
+class AlreadyReportCommentView(TemplateView):
```
maltem-za: Please rename to `AlreadyReportedCommentView`

gem/admin.py
```diff
@@ -17,5 +25,19 @@ def gender(self, obj):
...
 
+class GemReportCommentAdmin(MoloCommentAdmin):
+    inlines = (GemReportCommentModelAdmin,)
+    list_display = MoloCommentAdmin.list_display + ('reported_reason', )
+
+    def reported_reason(self, obj):
+        reported_for = GemReportComment.objects.filter(
+            comment=self.id
+        )
+        return reported_for
```
maltem-za: How does this show up? As links to the reports?

maltem-za:
+1

PR #120, opened 2016-07-12
Use latest Molo

smn:


smn:
@miltontony follow up PR. Tiny.

miltontony:
👍

PR #135, opened 2016-09-12
Fix the bug with Checkboxes and Radio Selects

saeedmb8:


saeedmb8:
@miltontony ready for review :)

saeedmb8:
👍  from @miltontony

PR #136, opened 2016-09-12
Feature/surveys tweaks

CSergienko:
@saeedmb8 @miltontony Please have a look and let me know if you're happy with this.

### DIFF
--- a/gem/static/css/_style.surveys.scss
+++ b/gem/static/css/_style.surveys.scss
@@ -114,6 +114,10 @@
           margin: 0 0 32px 0;
         }
     }
+
+    .call-to-action--surveys {
+        text-align: center;
+    }
 }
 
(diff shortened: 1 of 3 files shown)


gem/static/css/_style.surveys.scss
```diff
@@ -45,6 +45,85 @@
...
+        text-align: center;
+
+        .section-title {
+            text-align: left;
+
+          h2 {
+            height: 30px;
+            line-height: 30px;
+            margin: 0 0 24px 0;
+            font-family: 'AvenirNextCondensed-Medium', sans-serif;
```
markupguy: We can use a SASS var here much like `font-family: $font-family-avenirBold;`but not hte end of the world

markupguy:
+1

PR #152, opened 2016-10-06
made custom form input style sass placeholder selector more generic, …

CSergienko:
…applied it to the terms and conditions component

@markupguy Have a look and let me know if you're happy with this.


markupguy:
+1, Thanks Chris

PR #168, opened 2016-10-26
Update FED.md

Mitso:


Mitso:
@saeedmb8 please review

gem/templates/base.html
```diff
@@ -78,8 +78,13 @@
             {% endfor %}
         </div>
         {% endif %}
-        {% bannerpages %}
 
+        {% if 'ninyampinga.com' in request.get_host %}
+          <div class="banner__birthday">
```
saeedmb8: @Mitso can we add this in here https://github.com/praekelt/molo-gem/blob/develop/gem/templates/core/main.html#L22

saeedmb8:
👍

PR #197, opened 2017-01-23
Gemmolo 140:  Update profile view/edit page to conform to IoGT

Mitso:


Mitso:
@saeedmb8 @nathanbegbie @Ntu2koTrevor @miltontony 

Please review 

We need to update Molo.profiles for GEMMOLO, however the current version of profiles uses an extra field for gender and security questions are hardcoded amongst other things. 
Molo.profiles have got several upgrades since, please see urls.py, views.py, and base.py for the current molo.profiles implementation.

Please review and advise if we can upgrade.

Template Reg flow:

screenshot 2017-01-23 16 45 58

screenshot 2017-01-23 16 46 08

screenshot 2017-01-23 16 46 31

PR #198, opened 2017-01-24
GEMMOLO 199: Article comment list

Mitso:


Mitso:
@saeedmb8 @nathanbegbie please review.

@saeedmb8 on the Recommended Article list, what we were trying to do was that only the first 3 article layout as well as images show. While the rest will display on the article title and subtitle

screenshot 2017-02-16 16 46 34

screenshot 2017-02-16 16 54 38

Mitso:
@saeedmb8 @nathanbegbie thank you!
Can you please review the updated

gem

PR #210, opened 2017-02-15
Remove ability to delete index pages in Admin UI

nathanbegbie:


saeedmb8:
@nathanbegbie don't you need to bring any templates over?

saeedmb8:
👍

PR #213, opened 2017-02-17
Add social media & tags

Mitso:


Mitso:
@saeedmb8 please review 

on Footer Social media 
- The URL input field needs to strip out the HTTP/HTTPS protocol because when a user adds it the URL doesn't work because it it already added on the source link 
screenshot 2017-02-17 17 30 53

screenshot 2017-02-17 17 26 13

Mitso:
screenshot 2017-02-17 17 48 24

saeedmb8:
@Mitso good point on Footer Social media 🍰  I think lets remove the https from the link.
so we get the whole URL from what the user enters and just have `<a href="{{ site.value.link }}" ` I will make an issue on Molo to add a help text for it as well, but even if they give us `www.google.com` it would be fine.
other than that 👍 👍

PR #222, opened 2017-03-06
Update fonts folder path ref

Mitso:


Mitso:
@saeedmb8 please review fonts path update.

saeedmb8:
👍

PR #236, opened 2017-04-06
Create script to create recomended articles

nathanbegbie:


gem/migrations/0014_convert_recomended_articles.py
```diff
@@ -0,0 +1,77 @@
...
+    for article in articles:
+        stream_data = []
+        linked_articles = []
+        for block in article.body.stream_data:
+            if block['type'] == 'page':
+                try:
+                    linked_articles.append(ArticlePage.objects.get(
+                                    id=block['value']))
+                except:
+                    print(("[ERROR]: ArticlePage {} with id {} has "
```
miltontony: Instead of printing, you should log it as an error
```python
import logging
logging.error(...)
```

gem/migrations/0014_convert_recomended_articles.py
```diff
@@ -0,0 +1,77 @@
...
+
+from django.db import migrations
+
+
+def create_recomended_articles(main_article, article_list):
+    '''
+    Creates recommended article objects from article_list
+    and _prepends_ to existing recommended articles.
+    '''
+    from molo.core.models import ArticlePageRecommendedSections
```
miltontony: It's improper to import a model directly in migrations. 
Use `apps.get_model("yourappname", "Model")` instead.
```python
ArticlePageRecommendedSections = apps.get_model("core", "ArticlePageRecommendedSections")
```

gem/migrations/0014_convert_recomended_articles.py
```diff
@@ -0,0 +1,77 @@
...
+            ArticlePageRecommendedSections(
+                page=main_article,
+                recommended_article=article).save()
+
+
+def convert_articles(apps, schema_editor):
+    '''
+    Derived from https://github.com/wagtail/wagtail/issues/2110
+    '''
+    from molo.core.models import ArticlePage
```
miltontony: same as above, use `apps.get_model(...)`

gem/migrations/0014_convert_recomended_articles.py
```diff
@@ -0,0 +1,77 @@
...
+
+    for article in articles:
+        stream_data = []
+        linked_articles = []
+        for block in article.body.stream_data:
+            if block['type'] == 'page':
+                try:
+                    linked_articles.append(ArticlePage.objects.get(
+                                    id=block['value']))
+                except:
```
miltontony: What error are you expecting here? It's better to be explicit rather than a blanket catch all.

gem/migrations/0014_convert_recomended_articles.py
```diff
@@ -0,0 +1,77 @@
...
+    from molo.core.models import ArticlePage
+    from wagtail.wagtailcore.blocks import StreamValue
+
+    articles = ArticlePage.objects.all()
+
+    for article in articles:
+        stream_data = []
+        linked_articles = []
+        for block in article.body.stream_data:
+            if block['type'] == 'page':
```
miltontony: Have you thought about the following article layout and the implications of this migration:
```
Paragraph 1
Paragraph 2
Paragraph 3
Link to Page
Paragraph 4
Paragraph 5
Link to Page
Link to Page
```
screen shot 2017-04-08 at 1 13 25 pm
Also, removing the page link from the middle of an article will leave it in a bit of an un-readable state.

@saeedmb8 @nathanbegbie Please confirm that this is the expected behaviour with the content team.

PR #249, opened 2017-04-11
malawi header changes

projectmushroom:
@mitso please review 

Header bar now using the more compact version of the original GEM site, feature article image size back to rectangular dimension

Before:
screen shot 2017-04-11 at 5 05 01 pm

After:
screen shot 2017-04-11 at 5 04 24 pm


Mitso:
💯 🍰

PR #259, opened 2017-04-25
remove the unnecessarily settings

saeedmb8:


saeedmb8:
@Mitso quick review please

gem/settings/base.py
```diff
@@ -127,7 +127,7 @@
 
 # We have multiple layouts: use `base`, `malawi` or `springster`
 # to switch between them.
-SITE_LAYOUT_BASE = environ.get('SITE_LAYOUT_BASE', 'springster')
+SITE_LAYOUT_BASE = environ.get('SITE_LAYOUT_BASE', 'base')
```
Mitso: @saeedmb8 is changing this to base intentional?

Mitso:
@saeedmb8 otherwise a 👍

gem/settings/base.py
saeedmb8: @Mitso yes, for now on QA and prod the base site layout is gonna be the `base` and those that we want to be `springster` we can change the value of `SITE_LAYOUT_BASE` on MC

PR #264, opened 2017-05-02
Springster styling phase 2

projectmushroom:


gem/styles/gem-springster/modules/_m-banner.scss
```diff
@@ -1,6 +1,10 @@
 
 .banner__item-link {
   display: block;
+  img {
```
Mitso: We should add a class for banner images here to reuse styles as much as possible. 
My bad

gem/styles/gem-springster/modules/_m-tag-cloud.scss
```diff
@@ -0,0 +1,80 @@
...
+    position: absolute;
+    left: 50%;
+    margin-left: -50px;
+  }
+
+  &__heading {
+    @include zig-zag-underline;
+    text-align: center;
+
+    h4 {
```
Mitso: I saw the tag cloud had different variations on articles as well. On clean up we must be able to pick and choose pattern elements without loosing styles.

Mitso:
@projectmushroom just a few comments to keep in mind when we do a cleanup I think. Otherwise it's 💯 for land this.

gem/styles/gem-springster/modules/_m-tag-cloud.scss
projectmushroom: cool we can do these two in clean up, the buttons styles also needs some work

PR #267, opened 2017-05-05
Springster Redesign Phase 3

Mitso:


gem/styles/gem-springster/modules/tags/_m-tag-cloud.scss
```diff
@@ -64,17 +64,17 @@
 }
 
 // //Tag Cloud for Article Page, dont see the need of this?
```
Mitso: @projectmushroom this style tag is for article tag variation e.g. 
screenshot 2017-05-08 13 08 56

Mitso:
@projectmushroom you have updated the following button
 `.call-to-action__nav-item-text--right` and `.call-to-action__nav-item-text--left`
Please note the back [left] button looks like this for me 
screenshot 2017-05-08 13 40 24

projectmushroom:
@Mitso I'll sort out the button arrows, I like the  articles tags, nice 👍

gem/settings/base.py
```diff
@@ -127,7 +127,7 @@
 
 # We have multiple layouts: use `base`, `malawi` or `springster`
 # to switch between them.
-SITE_LAYOUT_BASE = environ.get('SITE_LAYOUT_BASE', 'base')
+SITE_LAYOUT_BASE = environ.get('SITE_LAYOUT_BASE', 'springster')
```
Mitso: We need to update this

Mitso:
💯 🍰

PR #273, opened 2017-05-09
Update ui elements for pages

projectmushroom:


gem/styles/gem-springster/_base.scss
```diff
@@ -1,8 +1,6 @@
 html, body  {
```
Mitso: I think the base font-size = 16px set here helps to using various size unit either %, em and etc.

Mitso:
@projectmushroom 💯 🍰 

👍  just one comment, I think we should keep base styles  on base file still.

gem/styles/gem-springster/_base.scss
projectmushroom: I've moved it to _m-typography.scss, also we're not using rems so adding it on the html tag seems unnecessary

PR #274, opened 2017-05-10
fix buttons

projectmushroom:
buttons were not displaying on non retina/HDPI devices, media queries removed


Mitso:
@projectmushroom 💯

PR #304, opened 2017-05-15
Comment out rtl lets for centered headings

Mitso:


Mitso:

PR #307, opened 2017-05-16
QA  Updates

Mitso:


Mitso:
There's currently no surveys results page design, made it similar to Polls results

screen shot 2017-05-16 at 12 24 41 pm

PR #311, opened 2017-05-17
Clean templates and remove spaces 

Mitso:


Mitso:

PR #319, opened 2017-05-18
Default input markup 

Mitso:


gem/styles/gem-springster/modules/buttons/_m-springster-buttons.scss
```diff
@@ -59,6 +58,7 @@
 }
 //The buttons must have a default color with no image loading for featurephones <=240px
 .call-to-action__button--primary {
+  background-color: $white;
```
projectmushroom: are we specifying a bg color to override the default user agent styles on inputs? if so, we should rather user background-color: transparent, we're using the buttons on other backgrounds as well, like eg. your words

projectmushroom: screen shot 2017-05-18 at 1 17 19 pm

Mitso: Ok cool, will update

PR #326, opened 2017-05-19
Update feature phone forms

Mitso:


gem/styles/gem-springster/modules/buttons/_m-springster-buttons.scss
```diff
@@ -65,8 +68,8 @@
   height: 40px;
 }
 .call-to-action__button--secondary {
-  background-color: #EBEEF0;
-  background-position: -1728px -292px;
+  background-color: #D7DBDD;
```
Mitso: That is the hex color of the secondary bg image. We need to set a default background color fallback on other devices

Mitso:
Updates

PR #328, opened 2017-05-22
Update profile banner styles

Mitso:


Mitso:

Moved the smart styles as state style. 
I'll discontinue working on this in the mean while

PR #329, opened 2017-05-22
Move carousel grid layout style to state

Mitso:


Mitso:

Update a better layout when grid is not fully supported



Fancy checkboxes and radio buttons not supported, can not find a better fallback solution here except for CSS media query 


Feature phone and slowend phone RTL and layout fixes

gem/styles/gem-springster/layout/_l-header.scss
```diff
@@ -3,7 +3,7 @@ html, body {
 }
 .header {
   overflow: auto;
-  padding: 10px 10px 5px;
+  padding: 10px 10px 0 5px;
```
projectmushroom: is this correct? bottom 0px and then right 10px and left 5px? initially it was set to have top, left and right 10px and 5px bottom

Mitso: @projectmushroom thanks. I've increased this size on smart style - Nokia phones a a white line break there unless bottom padding is 0 or 10px. 

I'm really not sure why.

PR #332, opened 2017-05-23
Search Toggle support 

Mitso:
…ch toggle


Mitso:
Article page comment counter update 
screen shot 2017-05-23 at 3 12 07 pm

PR #335, opened 2017-05-23
Tag cloud image - feature phone bug 

Mitso:


Mitso:

screen shot 2017-05-23 at 6 03 51 pm

PR #338, opened 2017-05-23
Footer image is not responsive feature phone

Mitso:


gem/styles/gem-springster/state/state.scss
```diff
@@ -17,6 +17,9 @@
       margin-left: 10px;
     }
   }
+  #footer-bar .footer-logo {
+    width: 200px;
```
projectmushroom: how does this effect the smartphone version? maybe we should add max-width:200px ?

Mitso: It actually needs a width otherwise it doesn't show the image.
I need to deploy to test this unfortunately

projectmushroom: oh okay, nice

PR #342, opened 2017-05-24
Html image alt tags update

Mitso:


Mitso:

PR #354, opened 2017-05-26
Section title and comment bubble rtl

Mitso:


gem/styles/gem-springster/state/comments/_s-comment-counter.scss
```diff
@@ -1,8 +1,10 @@
...
   font-weight: normal;
 }
-.comment-counter {
-  position: absolute;
-  margin-top: 0;
-  right: 10px;
+.teaser-stacklist__item--content {
+  .comment-counter {
+    position: absolute;
+    margin-top: 10px;
```
projectmushroom: how is this displaying on devices?

still aligned like this?
screen shot 2017-05-26 at 9 03 20 am

Mitso: @projectmushroom Yes it does. Thanks

PR #356, opened 2017-05-29
header fix

projectmushroom:
initial load content padding and transition font bug fix on webkit browsers


gem/styles/gem-springster/modules/menu/_m-menu.scss
```diff
@@ -72,6 +72,7 @@
     color: #fff;
     text-decoration: none;
     font-size: $font-small;
+    -webkit-font-smoothing: subpixel-antialiased;
```
Mitso: @projectmushroom this should be on state-smart/ state_320

Mitso:
@projectmushroom 💯 🍰 

Just to note 
I discovered 2 Issues on mobile & desktop browser.

On initial site load there is a space between banner and header until you visit a different page
screen shot 2017-05-29 at 10 29 05 am

On mobile content is behind header 
img_20170529_103353

projectmushroom:
@Mitso those are the problems that should now be fixed

PR #358, opened 2017-05-29
load more fixes + tag limit

projectmushroom:
- replaced the fetch API with a more traditional XMLHttpRequest, will now support older browsers
- `isMobile` added for user agent detection to prevent springster.js execution on feature phones, 
- base teaser tag's won't show if exceeding 20 characters


gem/static/js/springster.js
```diff
@@ -1,157 +1,170 @@
...
-      var errorList = form.querySelectorAll('.errorlist');
-      if (errorList.length > 0) {
-        for (var i = 0; i < errorList.length; i++) {
-          parent = errorList[i].parentNode;
-          parent.classList.add("input-error");
-        }
+            element.childNodes[1].innerHTML = "<img src='/static/img/loading.gif' alt='Loading...' />";
+            var meh = element.getAttribute('data-next');
+            getAjax(element.getAttribute('data-next'), function(data){
+              console.log(data);
```
Mitso: Do you need console.log() to production?

Mitso:
@projectmushroom looks good to me :100: :cake:

gem/static/js/springster.js
projectmushroom: 😲 thanks for spotting!

PR #359, opened 2017-05-30
hide load more on js friendly feature phones

projectmushroom:


gem/static/js/springster.js
```diff
@@ -7,6 +7,11 @@
       document.readyState === "interactive" || document.readyState === "complete" ? callback() : document.addEventListener("DOMContentLoaded", callback);
   };
 
+  var remNoJS = function() {
```
Mitso: @projectmushroom display:none; removes the element on the DOM, does this kick in before that?

Is there a difference between 
```
<noscript>
      <style>#more-link {display:none;}</style>
 </noscript>
```

// Toggle loadmore button on js freindly featurephones
```
.no-js .more-link {
  display: none;
}
```

projectmushroom: haha completely forgot about our little `<noscript>` not needed anymore

Mitso:
👍 💯

PR #365, opened 2017-05-30
Feature/springster fed bugs v7

projectmushroom:
- replace css capitalisation headings, with Django `capfirst` filter
- remove unused state from `springster.css`
- menu item wrap detection, reduce list container width to force down 2nd last menu item. (_this might be a bit experimental, but we'll need to test it on the UAT sites so see how it will react with various character widths and smartphone screen sizes_)
- js cleanup on forms


gem/static/js/springster.js
```diff
@@ -30,13 +30,20 @@
...
+    var menuList = document.getElementById("nav-list");
 
     var onResizing = function(event) {
-      if (window.innerWidth < 1024 ){
+      if (window.innerWidth < 1024){
         content.style.backgroundColor =  "#7300ff";
       } else {
         content.style.backgroundColor =  "transparent";
       }
+
```
Mitso: @projectmushroom Looks cool, so if the tags wrap, they will be removed?

Mitso:
@projectmushroom Nice Djnago capfirst find 🍰 
 The comment is testing once landed like you said. 👍

gem/static/js/springster.js
projectmushroom: this is just for the menu items currently, so when the menu wraps the last items, js pics up the altered height of the container, and gives it a smaller width to force down another item

_Homepage navigation - Can we force a second link down to the second line when it breaks. On e.g. the Spanish example the link Mi Vida is the only item on the second line_

PR #380, opened 2017-06-08
Update gem management commands

saeedmb8:


codiebeulaine:
@saeedmb8 I would prefer more tests on the success message and translations, but if that's unnecessary then 👍 (wtih)

PR #381, opened 2017-06-09
fix reaction Q template

projectmushroom:
error message messed up the page layout 

Before
screen shot 2017-06-09 at 2 34 24 pm

Fixed:
screen shot 2017-06-09 at 2 34 00 pm


Mitso:
:cake:

PR #387, opened 2017-06-13
ensure meta tags are in the templates

miltontony:


miltontony:
@codiebeulaine review please

PR #389, opened 2017-06-15
use manifest static file storage

miltontony:


miltontony:
@JayH5 looks like it's still failing 😒

JayH5:
_"Due to the requirement of running collectstatic, this storage typically shouldn’t be used when running tests as collectstatic isn’t run as part of the normal test setup. During testing, ensure that the STATICFILES_STORAGE setting is set to something else like 'django.contrib.staticfiles.storage.StaticFilesStorage' (the default)."_

https://docs.djangoproject.com/en/1.11/ref/contrib/staticfiles/#manifeststaticfilesstorage

miltontony:
ah.. I see in your PR it was working.. so I think it was fixed in `molo.core==5.5.0`

JayH5:
@miltontony yeah I moved the `ManifestStaticFilesStorage` settings to `settings/production.py`.

PR #392, opened 2017-06-21
update reaction question success image size

projectmushroom:
Before
screen shot 2017-06-21 at 3 20 35 pm

After
screen shot 2017-06-21 at 3 19 29 pm


Mitso:
🍰 👍

PR #397, opened 2017-06-30
Versions update

Mitso:


Mitso:
@projectmushroom  please review 

screen shot 2017-06-30 at 10 39 36 am

QA 
http://rogue.springster.seed.p16n.org/versions/

PR #408, opened 2017-07-11
Advanced Surveys

miltontony:


tm-kn:
FYI, waiting for LabD/wagtail-personalisation#169 to be merged and pushed to PyPI. We will need to bump version in the requirements file before merging.

tm-kn:
Wagtail-personalisation updated.

miltontony:
I'm going to land this into a branch, make some changes and merge to develop.

gem/personalise/models.py
```diff
@@ -0,0 +1,130 @@
...
+                label = '%s (%s)' %(label, field.segment.name)
+
+            data_fields.append((field.clean_name, label))
+
+        return data_fields
+
+    def serve(self, request, *args, **kwargs):
+        # We need request data in self.get_form_fields() to perform
+        # segmentation.
+        #TODO(tmkn): This is quite hacky, need to come up with better solution.
```
miltontony: @tm-kn What is the proper solution here?

tm-kn: Hi @miltontony, the solution would need to be provided upstream in [Wagtailsurveys](https://github.com/torchbox/wagtailsurveys). It would need a `HttpRequest` object to be passed into `get_form_fields` or upstream in Wagtail where we can access request object in the `Page` class. Currently we assign it to a class property whenever serve() is called, which does not seem like a good practice and feels quite hacky, but it works flawlessly.

PR #409, opened 2017-07-11
add teaser image min size

projectmushroom:
fix safari mac tag issue

screen shot 2017-07-11 at 2 42 45 pm


Mitso:
@projectmushroom :+1:  just a suggestion to add padding on the bottom so that the tags are not too tight

PR #410, opened 2017-07-14
translation issues

saeedmb8:


Mitso:
@saeedmb8 :cake: from me.

Just a question, why are we removing some already existing translation e.g. 
gem/templates/comments/comment_block and gem/templates/wagtailadmin/pages/listing/_page_title_explore

saeedmb8:
@Mitso I shouldn't have remove the `comment_block` one thanks for noticing that 🍰 but we don't need to translate the admin texts, we only translate the words that users see and those have been added to pootle accidentally.

PR #413, opened 2017-07-17
add malawi homepage fb share image

projectmushroom:


Mitso:
@projectmushroom the {{request.site.root_url}} or root URL image is already defied on the else statement.  Maybe you can feed this Zathu static image there?

projectmushroom:
@Mitso sure NP, was trying to avoid the extended length inline statement, updated

PR #418, opened 2017-07-25
adding latest commenting (5.1.0)

codiebeulaine:
This update is needed so that we can make use of the email csv option on gem as the current download as csv option is timing out. This change was made in commenting 5.1.0


cblignaut:
+1

PR #420, opened 2017-07-31
Feature/add latest surveys for convert article

codiebeulaine:
This PR upgrades to the latest surveys which has convert to article functionality as wel as custom submit survey message and results as percentage. 

The latest version of surveys does also include the ability to display a survey directly, however the front end work for this will need to happen in a separate PR.


saeedmb8:
👍

PR #425, opened 2017-08-08
Add GA Account for certain URL

nathanbegbie:
This PR does 2 things:
- Creates an additional GEM specific setting that allow for an additional Google Analytics tracking ID to be added to the site
- Replaces the custom `molo.core` middleware layer that handles google analytics traffic. If the request URL begins with `bbm` and the site contains the additional GA tracking code, the GA data is diverted to the additional GA account.


nathanbegbie:
Link to video demo: https://drive.google.com/a/praekelt.org/file/d/0B-xXQSFI0MakcFBMc2tldWJKYzA/view?usp=sharing

nathanbegbie:
@saeedmb8 please review

gem/migrations/0018_add_extra_ga_account.py
```diff
@@ -0,0 +1,30 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('gem', '0017_show_join_banner_setting'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='gemsettings',
+            name='extra_ga_account',
```
miltontony: I think you'll need to be explicit that this is for BBM. Especially since the business logic under the hood looks at this field and the subdomain one. 

I would recommend "BBM GA Account" and "BBM Subdomain"

PR #445, opened 2017-09-14
enable template caching in production

miltontony:
Wagtail recommends enabling template caching to improve performance.
http://docs.wagtail.io/en/v1.9/advanced_topics/performance.html#templates


miltontony:
@smn @saeedmb8 first PR for performance improvements.

codiebeulaine:
@miltontony just curious, what does template caching mean? Does it mean we will only load content in prod and the template would load once and be cached?

codiebeulaine:
👍

smn:
@miltontony what does this cacher do? Saves the read from disk? I'm not sure it caches anything relating to DB queries though.

miltontony:
@smn no, it doesn't do DB caching. Just disk.

smn:
:+1:

PR #460, opened 2017-09-28
update media components

projectmushroom:


gem/templates/base/core/article_page.html
```diff
@@ -33,8 +33,8 @@ <h4 class="title">{{media.title}}</h4>
                 <div class="media-player__embed-object">
                   {% embed media.youtube_link %}
                 </div>
-                <a href="{{media.file.ur}}" class="btn">
-                  <span class="call-to-action__button button">Download</span>
+                <a href="{{media.file.ur}}" class="btn" download>
```
miltontony: What's the `download` attribute here?

miltontony: Didn't know about this.. now I know :)

miltontony: https://www.w3schools.com/TAgs/att_a_download.asp for anyone else

PR #464, opened 2017-10-03
Add Facebook Analytics

nathanbegbie:
Ticket: https://praekeltorg.atlassian.net/browse/GEMMOLO-448

I have also moved other tags that result in `<script>` tags, to the bottom of our web page, in order to reduce page render time.

### DIFF
--- a/gem/templates/base/base.html
+++ b/gem/templates/base/base.html
@@ -39,30 +39,6 @@
     </head>
 
     <body class="{% block body_class %}{% endblock %}" {% if LANGUAGE_CODE|language_bidi == True %}dir="rtl"{% endif %}>
-        {% if settings.core.SiteSettings.ga_tag_manager %}
-        <!-- Google Tag Manager -->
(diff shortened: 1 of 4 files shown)


Mitso:
🍰 @nathanbegbie 

I'm not too sure about the bbm analytics though.

gem/templates/base/base.html
```diff
@@ -152,5 +128,32 @@
         </div>
 
         {% wagtailuserbar %}
+        {% block facebook_analytics %}
+          {% include "core/facebook_analytics.html" %}
+        {% endblock %}
+
+        {% if settings.core.SiteSettings.ga_tag_manager %}
+        <!-- Google Tag Manager -->
+         <noscript><iframe src="//www.googletagmanager.com/ns.html?id={{settings.core.SiteSettings.ga_tag_manager}}&client_session_id={{request.session.MOLO_GA_SESSION_FOR_NOSCRIPT}}&page_title={% if self.seo_title %}{{ self.seo_title|urlencode }}{% else %}{{ self.title|urlencode }}{% endif %}{%if request.META.HTTP_X_IORG_FBS_UIP %}&freebasics_uip={{request.META.HTTP_X_IORG_FBS_UIP}}{%endif%}"
```
projectmushroom: are the iframes working with freebasics? _"Some suppression may occur for iframes so make sure that your content doesn’t rely on them. We recommend avoiding iframes altogether as it tends to lessen a mobile experience."_  https://developers.facebook.com/docs/internet-org/platform-technical-guidelines/

codiebeulaine: @projectmushroom we don't use tag manager for springster for now so this code wont execute because of the if statement

PR #468, opened 2017-10-10
Springster favicon

Mitso:


Mitso:
Please review 
skitch

gem/templates/springster/base.html
```diff
@@ -30,6 +30,16 @@
     <meta property="og:image:width" content="450" />
     <meta property="og:image:height" content="200" />
     <meta name="viewport" content="width=device-width, initial-scale=1, {% if 'bbm.' in request.get_host %}user-scalable=0{% endif %}"/>
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
```
nathanbegbie: @Mitso does this not just repeat the line above it? - the BBM stuff will only get added if it's via the BBM viewport

nathanbegbie:
@Mitso 1 comment, otherwise +1 from me

Mitso:
That is indeed the case. Thanks @nathanbegbie

PR #469, opened 2017-10-10
Loader gif

Mitso:


gulpfile.js
```diff
@@ -78,4 +78,4 @@ gulp.task('watch', function() {
 });
 
 gulp.task('styles', ['styles:dev', 'styles:prd','stylesAdmin']);
-gulp.task('default', ['styles', 'watch']);
+gulp.task('default', ['styles','compress']);
```
miltontony: Why the change here? I don't see how this is related to the loader? Just curious..

nathanbegbie:
@Mitso Could we put a GIF in here of how the spinner behaves?

gulpfile.js
Mitso: The JS is not included in the default tasks, will updated ['styles','compress','watch']

Mitso:
Looks a bit tiresome

laoder

Mitso:
@miltontony @nathanbegbie please review 

The ticket requires a spinner or progress bar. 
Decided to implement at a progress bar instead, it looks to be functioning better than a spinner however 2 issues with a page loader. 
1. it's implemented only after the DOM is loaded 
2. Is annoying to me at least - will potentially be a UX drop-off issue

I've throttled performance here:
laoder

PR #472, opened 2017-10-12
Update tags template pattern

Mitso:


Mitso:
@nathanbegbie @saeedmb8 please review.

Tag pattern cleanup
Profiles word-break style 
Primary buttons loader js 

buttons-load

Mitso:

gem/static/js/springster.js
```diff
@@ -112,42 +112,26 @@
...
-    }
-    iterateButtonsOnPage(ctaButtonPrimary);
-
+  var loaderAnimation = function() {
+    var wrapper = document.getElementById("content-wrapper")
+    
+    wrapper.addEventListener("click", function(event){
+      var element = event.target;
+      if (element.classList.contains("call-to-action__button--primary")) {
+        element.setAttribute('value', '');
```
Mitso: Is this necessary then? Doesn't seem to clear out the value, I don't think it actually should to be honest.

PR #477, opened 2017-10-19
js updates

projectmushroom:
Had some issues on Safari where the loader didn't show, refactored the click behaviours of the buttons


gem/static/js/springster.js
```diff
@@ -113,15 +113,14 @@
...
-    wrapper.addEventListener("click", function(event){
-      var element = event.target;
-      if (element.classList.contains("call-to-action__button--primary")) {
-        element.setAttribute('value', '');
-        element.innerHTML = '';
-        element.classList.add('call-to-action__button--loader');
-       }
+    var ctaBtn = document.querySelectorAll('.call-to-action__button--primary');
+
+    Array.from(ctaBtn).forEach(link => {
```
nathanbegbie: Are we transpiling our JS? Just because `=>` may not work for some older browsers.

nathanbegbie:
:+1: from me (as long as we are def transpiling our JS)

PR #478, opened 2017-10-24
Adjustment of section name of articles

projectmushroom:
When an article is promoted in the latest section and it has no image on, then the section name hovers incorrectly over the banner / content on the top of the site

**Before:**
screen shot 2017-10-20 at 3 33 59 pm
incorrect option 1

**After:**
screen shot 2017-10-24 at 11 18 46
screen shot 2017-10-24 at 11 19 25


Mitso:
🍰 🍰 Looks good

PR #480, opened 2017-10-31
add JS for download event

codiebeulaine:


codiebeulaine:
@saeedmb8 @projectmushroom ready for review

gem/templates/malawi/core/article_page.html
```diff
@@ -47,7 +47,7 @@ <h4 class="title">{{media.title}}</h4>
                   <source src={{block.value.file.url}} type="video/mp4">
                   Your browser does not support the video tag.
               </video>
-              <a href="{{block.value.file.url}}" class="btn" download>
+              <a href="{{block.value.file.url}}" onclick=”var that=this;_gaq.push([‘_trackEvent’,’Download’,’Video’,this.href]);setTimeout(function(){location.href=that.href;},200);return false;” class="btn" download>
```
nathanbegbie: Do we not need to wrap this in some logic so that it doesn't show up if the user gets here on Free Basics?

codiebeulaine: @nathanbegbie that logic exists in this template a little above this on line 43

PR #485, opened 2017-11-02
add fb pixel tracker

codiebeulaine:


nathanbegbie:
Do Malawi and Rwanda need fbook tracking as well? If not, +1 from me

codiebeulaine:
@nathanbegbie there is no tracking code for them

PR #493, opened 2017-11-09
Fix the issue with mobile number error not showing up

saeedmb8:


saeedmb8:
Thanks @projectmushroom, it's been showing up in the GA page views result, so we usually ignore the path that we don't want to be tracked in GA

PR #501, opened 2017-11-14
GEMMOLO 534: Tag cloud update

Mitso:


Mitso:
Add span between anchor :before natural space

PR #504, opened 2017-11-16
Gemmolo 536: Advsurveys formfields label case

Mitso:


Mitso:

PR #510, opened 2017-11-22
Fix advanced surveys bugs

nathanbegbie:


alexmuller:
Pushed a commit for the survey pagination but disappointingly I've really struggled to write a view test for it today. There's a lot of setup which I can't get quite right.

PR #515, opened 2017-11-23
update form select error surfaced on required

Mitso:


Mitso:
Required select was not submiting upddated.
screen shot 2017-11-23 at 7 39 51 pm

Checks default checkbox, radio values

PR #524, opened 2017-12-01
Ensure SECRET_KEY is set correctly

alexmuller:
While looking through the codebase for the first time this setting worried me a little bit. In my experience it's very easy to forget to set this environment variable, which will cause the SECRET_KEY to be disclosed.

This commit means the app refuses to start unless the SECRET_KEY is set correctly. This will break currently running applications if they are upgraded to include this commit.


alexmuller:
Hm, having some trouble with the way this interacts while building containers with Docker. Going to close this for now and have a think about it.

PR #528, opened 2017-12-06
Profiles update - traceback

Mitso:


Mitso:
@saeedmb8 @projectmushroom please review 

2 issues I picked up:

on ViewProfile page
The setting option of `capture_something_on_reg` option is not accessible by some fields e.g. `settings.profiles.UserProfilesSettings.capture_location_on_reg`

Access registered input value with e.g. `user.gem_profile.get_gender_display` instead of updated `user.profile.gender`

This doesn't update both from when enabled on Registration or when Editing Profile page

codiebeulaine:
👍

PR #531, opened 2017-12-08
Backbutton bug

Mitso:


Mitso:
Added HTML  `novalidate` to remove browser default inline popup error validation which doesn't fire Django error validation while JS button spin is triggered and hangs.

saeedmb8:
👍

PR #534, opened 2017-12-08
add fb banner

codiebeulaine:


codiebeulaine:
screen shot 2017-12-08 at 2 58 29 pm

gem/settings/base.py
```diff
@@ -133,7 +133,7 @@
 
 # We have multiple layouts: use `base`, `malawi` or `springster`
 # to switch between them.
-SITE_LAYOUT_BASE = environ.get('SITE_LAYOUT_BASE', 'base')
+SITE_LAYOUT_BASE = environ.get('SITE_LAYOUT_BASE', 'springster')
```
saeedmb8: @codiebeulaine please change this back to base

gem/migrations/0022_auto_20171208_1336.py
```diff
@@ -0,0 +1,33 @@
+# -*- coding: utf-8 -*-
```
saeedmb8: can you please give the migration a name

gem/templatetags/gem_tags.py
```diff
@@ -14,6 +16,31 @@ def fieldtype(field):
...
+
+    if request.site:
+        pages = request.site.root_page.specific.bannerpages().exact_type(
+            GemTextBanner)
+    else:
+        pages = []
+    return {
+        'bannerpages': get_pages(context, pages, locale),
+        'request': context['request'],
+        'locale_code': locale,
```
saeedmb8: Suggestion: if we are going to have a template tag for the banner you could just pass in the `is_via_freebasics: context['is_via_freebasics']` and not have the `is_via_freebasics` template tag. but happy to have the template tag if you prefer it that way

saeedmb8:
left a suggestion there but if you happy with the current approach then 👍 🍰

PR #535, opened 2017-12-11
Upgrade molo.surveys to 6.1.1

alexmuller:
https://github.com/praekelt/molo.surveys/compare/6.1.0...6.1.1

This will cause some additional database reads for logged-in users.

This pull request includes a separate unrelated commit to urlpatterns which should have no actual change in behaviour.


codiebeulaine:
👍

PR #536, opened 2017-12-13
[SAS-67] Add date_joined field to admin email export

alexmuller:
This field is required for the Girl Effect dashboard.


alexmuller:
Thanks! Somebody once told me about applying the the scouts rule of [leaving the campsite cleaner than you found it](http://pragmaticcraftsman.com/2011/03/the-boy-scout-rule/) to code and I think that's a nice analogy :)

PR #537, opened 2017-12-14
add font lang fallback mixin

projectmushroom:
SCSS font fallback mixin added for languages not supported by the main Springster font, `Baton`
checks for language code in root element


Mitso:
@projectmushroom looks great 🍰 🍰 👍

PR #541, opened 2017-12-18
Back to top button fix

projectmushroom:
1. Revert back to top scroll behaviour (We previously used a flexbox layout that prevented the traditional back to top behaviour, required a some JS, now removed)
2. Fix back to top pattern alignment and height

Before:
screen shot 2017-12-18 at 16 39 53

After:
screen shot 2017-12-18 at 16 40 06


nathanbegbie:
:+1:

PR #546, opened 2018-01-08
add comment report reasons to model admin

codiebeulaine:


codiebeulaine:
screen shot 2018-01-08 at 3 46 02 pm

gem/admin.py
```diff
@@ -285,6 +286,26 @@ def gender(self, obj):
...
 
+class GemCommentModelAdmin(MoloCommentsModelAdmin):
+    list_display = (
+        'comment', 'moderator_reply', 'content', '_user', 'is_removed',
+        'is_reported', 'reported_count', 'reported_reason', 'submit_date')
+
+    def reported_reason(self, obj):
+        all_reported_reasons = []
+        for report in GemCommentReport.objects.filter(comment=obj.pk):
+            all_reported_reasons.append(report.reported_reason)
```
alexmuller: Instead of the loop I was curious if there was another way of doing this. Turns out there's a `values_list()` but I'm not sure it's really nicer:

```
all_reported_reasons = list(GemCommentReport.objects.filter(comment=obj.pk).values_list('reported_reason', flat=True))
```

PR #550, opened 2018-01-10
Module version update install

Mitso:


Mitso:
I also don't know enough tbh but from what I can see it changes between npm  packages yes. The toplevel  _marked_ module now runs on v0.3.9 however other npm packages like marked-terminal, cli-usage package-lock.json  references the vulnerable version v0.3.6. The above solve seems to fix the vulnerability issue.

PR #557, opened 2018-01-16
Tidy before supporting Python 3

alexmuller:


alexmuller:
@codiebeulaine spotted that the `LogHeaderInformationMiddleware` isn't used anymore 👏

PR #560, opened 2018-01-18
declare ga for download events

codiebeulaine:


nathanbegbie:
@codiebeulaine why not pop this script into the bottom of the `base.html` file?

PR #562, opened 2018-01-18
Render blocking optimisation 

Mitso:


gem/settings/base.py
```diff
@@ -133,7 +133,7 @@
 
 # We have multiple layouts: use `base`, `malawi` or `springster`
 # to switch between them.
-SITE_LAYOUT_BASE = environ.get('SITE_LAYOUT_BASE', 'base')
+SITE_LAYOUT_BASE = environ.get('SITE_LAYOUT_BASE', 'springster')
```
nathanbegbie: Just be careful with this as we don't want to commit it.
It's safer run `export SITE_LAYOUT_2='springster'` in the terminal window, then `./manage.py runserver`.

PR #571, opened 2018-01-24
Template optimizations

Mitso:


codiebeulaine:
@Mitso 👍 please add before and after stats of how fast the site runs to the ticket

PR #572, opened 2018-01-25
Add migration for security questions

alexmuller:
This commit creates security questions for each of the settings.

It applies to each of the sites in a multi-site setup and creates translations in each language on the site.

I haven't written any tests for this and I don't know if I should. It feels kind of tricky to write tests for a migration like this.


alexmuller:
I guess coveralls wants me to write tests…

codiebeulaine:
🚬

codiebeulaine:
🍻

PR #574, opened 2018-01-26
Test if user matches static segmentation rules even without an active request

KaitCrawford:
Currently you can only test if a user matches a rule if they have an active session. This change allows us to also test using a User from the database.


saeedmb8:
👍

saeedmb8:
🍰

PR #575, opened 2018-01-26
Enable yourwords survey functionality as GCN

Mitso:


Mitso:

Survey - Is Yourwords competition setting enabled
screen shot 2018-01-30 at 15 30 00

Surveys - Polls or no setting enabled default functionality.
screen shot 2018-01-30 at 15 30 18
screen shot 2018-01-30 at 15 30 41

Mitso:
Homepage position

screenshot-2018-1-30 main

gem/templates/springster/core/main.html
```diff
@@ -10,7 +10,7 @@
     {% media_listing_homepage %}
   {% else %}
     {% get_tag_articles tag_count=4 as tag_nav_data %}
-    {% paginate 12 tag_nav_data.latest_articles as latest_articles %}
+    {% paginate 15 tag_nav_data.latest_articles as latest_articles %}
```
projectmushroom: I wonder what the homepage loading impact will be with less articles. We were already pushing it with 12, now we are on 15, the home page is getting quite big, we should really rework this after our optimisation investigations.

PR #577, opened 2018-01-26
use molo profiles migrated username instead of gem migrated username

codiebeulaine:


gem/settings/base.py
```diff
@@ -427,7 +427,7 @@
 
 
 AUTHENTICATION_BACKENDS = [
-    'gem.backends.GemModelBackend',
+    'molo.profiles.backends.MoloProfilesModelBackend',
```
alexmuller: This change needs to happen in `gem/settings/production.py` as well

alexmuller: Wait actually why do we need to set this in `production.py` at all? It's exactly the same as here.

PR #579, opened 2018-01-26
remove migrated username from gem profile

codiebeulaine:


alexmuller:
#577 needs to be in production before we merge this, right?

PR #582, opened 2018-01-29
remove gem gender usage

codiebeulaine:
TODO:: make sure gender field is shown for molo profiles admin and csv emails


alexmuller:
It looks like the `GemUserAdmin` is removing an admin view that includes gender and I can't see where that is now in the admin. When I run this locally I can't see the gender field but it is there on QA.

PR #591, opened 2018-02-02
use molo profiles settings for registration, edit profile, and view p…

codiebeulaine:
…rofile


codiebeulaine:
cc @Mitso @projectmushroom

PR #593, opened 2018-02-06
Lock node modules dependancies

Mitso:


Mitso:
@projectmushroom re view please.
Removed, cleaned up  dependencies

PR #604, opened 2018-02-11
add .specific to pageurl

codiebeulaine:


codiebeulaine:
@alexmuller once this is part of wagtail's code we won't need to say specific. But very much noted

PR #605, opened 2018-02-12
Ger 22 update long buttons

Mitso:


Mitso:

screen shot 2018-02-12 at 12 17 45

PR #607, opened 2018-02-12
concatenate strings correctly

codiebeulaine:


codiebeulaine:
screen shot 2018-02-12 at 11 34 51 pm

alexmuller:
I think there might be a problem here with the order of the HTML fieldset and the Django template tags?

```
{% for question, answers in results.items %}
  <fieldset>
    {% for answer, count in answers.items %}
    </fieldset>
  {% endfor %}
{% endfor %}
```

PR #610, opened 2018-02-14
Add template lint script

alexmuller:


alexmuller:
This will catch templates like this:

```jinja
{% load core_tags core_tags %}
```

And also this:

```jinja
{% load core_tags %}
<h1>core_tags not needed here</h1>
```

nathanbegbie:
@alexmuller do you think we could pop `template_lint.py` in `molo.core` so that it's more reusable?

alexmuller:
That is technically a much better idea but will need a couple of tweaks, please hold the line.

I'm kind of surprised I can't find a better one on PyPI that someone else has written, ideally I'd like to publish it there but can't be bothered / too much work right now.

alexmuller:
Can we get this merged in? There are quite a few template errors on the develop branch which I've fixed in #711, for example:

- `templates/springster/core/tags/load_tags_for_homepage.html` contains `endblock` but no `block`
- `templates/springster/core/tags/nav_tags_section_listing_homepage.html` contains `image` tag which isn't loaded
- `templates/springster/patterns/components/article-page/sp_variations/main-article-page.html` contains `bbm_share_url` tag which isn't loaded
- `templates/springster/registration/password_reset_done.html` contains `pageurl` tag which isn't loaded

codiebeulaine:
👍 wtih

PR #612, opened 2018-02-14
add survey update with homepage introduction

codiebeulaine:


codiebeulaine:
👍 on the template work @Mitso

PR #613, opened 2018-02-15
Introduce template fragment caching

alexmuller:
It's hard to measure exactly the impact of this because there are currently so many database queries for the homepage. In development I need to remove the final `{% get_pages %}` to get the page to load successfully with the debug toolbar.

- `develop`: 2801 queries in 1537.64ms
- this branch with a warm cache: 1369 queries in 845.82ms

Before merging this I'd like to make sure we have some good content and measurements for QA and then we'll need to test this thoroughly by publishing new content and observing the result.

## Cache header and footer template for 1 hour 

I think these could probably be longer because I doubt they change much but an hour is fine.

This removes about 250 SQL queries from a homepage load.

## Cache homepage banners for 15 minutes

My understanding is that this means new banners could take up to 15 minutes to appear on the homepage.

We could probably make this smarter by removing the cached items when a new banner is published, but for the time being I think this basic approach is fine.

### DIFF
--- a/gem/templates/springster/base.html
+++ b/gem/templates/springster/base.html
@@ -1,4 +1,4 @@
-{% load compress wagtailuserbar core_tags staticfiles %}
+{% load cache compress wagtailuserbar core_tags staticfiles %}
 {% load wagtailsettings_tags wagtailimages_tags molo_commenting_tags %}
 {% get_settings %}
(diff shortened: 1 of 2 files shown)


alexmuller:
I'll give a bit of context on how this works if that helps:

```

PR #616, opened 2018-02-16
Retrieve user info from rules

KaitCrawford:
In order to show how a user matched the rules of a segment we need to get the user's data for each rule. Only the rule knows which field must be retrieved and what it should be called.
These methods will be called when compiling a csv of the users that matched a static segments rules.

This adds methods for getting that data to the Profile Data and Comment Count segmentation rules.


KaitCrawford:
@alexmuller Yeah, if this's going away then that's definitely where this should have been done :(
I wasn't aware of that. I'll create an issue here and maybe talk to Lisa about those plans

PR #620, opened 2018-02-16
Add dev logging

alexmuller:
There are lots of exceptions from templates so we should make those visible in the development logs.


sewagodimo:
it looks good to go :cake:

PR #621, opened 2018-02-19
Reverse pwa version

Mitso:


Mitso:

Revert PWA for current error
`'WSGIRequest' object has no attribute '__getitem__'`
`user_is_authenticated` tag

PR #622, opened 2018-02-19
Fix checkboxes not having unique IDs

nathanbegbie:
`field.id_for_label` elvaluates to None in the form template. This means that checkboxes ended up with the same id, resulting in incorrect behaviour when checking boxes.

This is a temporary solution and warrants further investigation as to why this particular method fails.

The reason for the custom rendering of the form field, is so that the `<input>` is alongside the `<label>` element, rather than nested inside it, so that it can be styled appropriately.

kapture 2018-02-19 at 12 07 50


nathanbegbie:
After fix:
kapture 2018-02-19 at 12 10 45

Mitso:
@nathanbegbie Looks great to me 🍰 👍 
Will need another back-end  dev to review of course.

codiebeulaine:
👍 well done Nathan!!

PR #623, opened 2018-02-19
Compress smart styles with django hashing

Mitso:
…g of files


Mitso:
@projectmushroom a review please. 
Cleanup images dir and css defer

PR #624, opened 2018-02-20
combine teaser and tag patterns

projectmushroom:


Mitso:
@projectmushroom 🍰 🍰 👍

PR #625, opened 2018-02-20
Update checkbox labels across

Mitso:


gem/settings/base.py
```diff
@@ -132,7 +132,7 @@
 
 # We have multiple layouts: use `base`, `malawi` or `springster`
 # to switch between them.
-SITE_LAYOUT_BASE = environ.get('SITE_LAYOUT_BASE', 'base')
+SITE_LAYOUT_BASE = environ.get('SITE_LAYOUT_BASE', 'springster')
```
codiebeulaine: can we remove this change please

codiebeulaine:
👍 wtih

Mitso:
@codiebeulaine thanks for review.

Just a note this fix is only for working for Surveys, Polls  that are checkboxes and Reaction  question do not pull Question ID.

PR #626, opened 2018-02-20
Remove template caching from header

alexmuller:
This depends on request.GET and request.path which will change based on the path.

We can probably do this better somehow but I can't think of exactly how just yet.


codiebeulaine:
👍 wtih

PR #640, opened 2018-02-26
Ensure secret key is set correctly

alexmuller:
This raises an error when the application starts if the secret key is not set correctly.

### DIFF
--- a/gem/tests/wagtailsearch/test_wagtailsearch_base.py
+++ b/gem/tests/wagtailsearch/test_wagtailsearch_base.py
@@ -1,0 +1,9 @@
+from django.test import TestCase
+
+from gem.wagtailsearch.backends.base import BaseSearchBackend
+
+
+class TestWagtailSearchBase(TestCase):
+    def test_base_search_backend_get_index_for_model_returns_none(self):
+        backend = BaseSearchBackend({})
+        self.assertEqual(backend.get_index_for_model(None), None)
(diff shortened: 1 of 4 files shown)


gem/tests/wagtailsearch/test_wagtailsearch_base.py
```diff
@@ -0,0 +1,9 @@
+from django.test import TestCase
+
+from gem.wagtailsearch.backends.base import BaseSearchBackend
+
+
+class TestWagtailSearchBase(TestCase):
+    def test_base_search_backend_get_index_for_model_returns_none(self):
+        backend = BaseSearchBackend({})
+        self.assertEqual(backend.get_index_for_model(None), None)
```
nathanbegbie: @alexmuller can you please explain what this test is supposed to do or how it's relevant to the secret key?

alexmuller: @nathanbegbie Yeah sorry, it's nothing to do with the secret key. It's just a very simple test which means that we're at least making sure the `wagtailsearch` code executes. I only did it because of the angry red cross from coveralls on the previous commit (it's hard to test production settings), no other reason.

nathanbegbie: Ha ha, cool! Thanks for the explanation - I think seaworthy might be what this is for? Not sure where we'd put those tests though. Anyway, can't be solved in this PR, so happy to approve.

PR #643, opened 2018-03-05
Added anonymous commenting settings to molo-gem

sewagodimo:
Added the anonymous settings to molo commenting views. Insteading of defaulting to "anonymous" when non-staff users comment anonymously., the html now retrieves a value from settings.


sewagodimo:
Hey @codiebeulaine @KaitCrawford @miltontony
 I added the tests so check if the anonymous commenting alias is being translated. I have also update the requirements for molo.commenting to retrieve the most recent version of molo.commenting. I then updated .travis.yml to include all the `./manage.py compilemessages' instruction so that the translations do not fail. Whoop whoop!

.travis.yml
```diff
@@ -12,6 +12,8 @@ install:
   - pip install coveralls
 script:
   - flake8
+  - python manage.py makemessages -l es
```
codiebeulaine: perhaps let's makemessages for all languages and not es

codiebeulaine:
@sewagodimo we need to add tests here for the the customised anonymous alias, also let's have a chat with @saeedmb8 about the translations.

.travis.yml
```diff
@@ -12,6 +12,8 @@ install:
   - pip install coveralls
 script:
   - flake8
+  - python manage.py makemessages
+  - python manage.py compilemessages
```
saeedmb8: we shouldn't add these in travis file, the command makemessages should always be done locally, also we already compile the messages in the [docker file ](https://github.com/praekelt/molo-gem/blob/develop/Dockerfile#L7)

gem/tests/test_views.py
```diff
@@ -431,6 +440,23 @@ def test_comment_filters(self):
 
         self.assertFalse(comment_form.is_valid())
 
+    def test_anonymous_comment_translation(self):
```
saeedmb8: we don't need to test for pootle translations as they often get changed. so I don't think this is necessarily

gem/templates/base/comments/comment.html
```diff
@@ -14,10 +14,10 @@
           {% trans "Big Sister" %}
         {% endif %}
     {% elif node.user_name.lower == 'anonymous' %}
-      {{node.user_name}}
+      {% trans settings.commenting.CommentingSettings.commenting_anonymous %}
```
saeedmb8: we are not able to translate a value of variable that is not set in the template. For Pootle you would need the text that you want to be translated in the template. so the above syntax is not right.

PR #645, opened 2018-03-05
Polls via surveys

Mitso:


Mitso:

Polls via Survey Springster:

Homepage direct form:
screen shot 2018-03-06 at 11 45 32

Results page and results with percentage setting:
screen shot 2018-03-06 at 12 01 48
screen shot 2018-03-06 at 11 58 53

PR #649, opened 2018-03-06
Add polls via surveys on base templates

Mitso:


Mitso:

Customize Polls, Yourwords via Surveys
screen shot 2018-03-07 at 13 14 51

screen shot 2018-03-07 at 13 15 02

screen shot 2018-03-07 at 13 15 11

PR #652, opened 2018-03-07
Remove unwanted registration fields

alexmuller:
Girl Effect never want to collect email addresses, location or mobile numbers during registration, so we should remove them from the form to make it impossible for an accidental misconfiguration to result in them being collected.

https://github.com/praekelt/molo/compare/6.7.4...6.7.5


saeedmb8:
👍

PR #656, opened 2018-03-08
Clean up survey templates

Mitso:


Mitso:

Cleanup and fix on Zathu

Issue Polls via surveys:
screen shot 2018-03-09 at 10 52 41

Update:
screen shot 2018-03-09 at 10 51 35

PR #659, opened 2018-03-12
Update advance survey skip logic links

Mitso:


Mitso:
Update template logic as markup gets complicated. 

Part 2 - Remove duplicate styles files for Malawi and inherit from Base.

gem/styles/gem-malawi/modules/_m-your-words.scss
```diff
@@ -1,56 +1,8 @@
...
-// Check other forms, we may be able to share styles.
 .your-words {
-  background: #ae016d;
-  color: #fff;
-  text-align: center;
-
   form {
-    margin: 0 8px;
-    text-align: left;
+    //Make Note
```
projectmushroom: okay, noted... 😀

Mitso: Oh yeah! :D I shall delete now

PR #661, opened 2018-03-12
Refactor BBM logic into context processor

alexmuller:
Each of these `request.get_host` in the template is a call to the `get_host()` method on a `WSGIRequest`. It's hard to tell because in the template syntax we can call a method without the parentheses.

This probably isn't a massive performance problem but it's unnecessary. It's better to group this logic together so that we only run it once in the context processor. This also means we can test it easily.

### DIFF
--- a/gem/context_processors.py
+++ b/gem/context_processors.py
@@ -1,6 +1,12 @@
 from django.conf import settings
 
 
+def detect_bbm(request):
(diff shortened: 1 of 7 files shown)


projectmushroom:
nice, context processor FTW ! 👍🏼🍰🍰

PR #663, opened 2018-03-12
[GEMMOLO-840] Set ALLOWED_HOSTS correctly

alexmuller:
This is a Django security feature which allows use to protect against HTTP Host header attacks.

https://docs.djangoproject.com/en/1.11/ref/settings/#std:setting-ALLOWED_HOSTS

Instead of setting it to a wildcard we should set it to the hosts we expect to be running this application.

Domains starting with a dot are wildcard subdomains.

### DIFF
--- a/gem/settings/base.py
+++ b/gem/settings/base.py
@@ -32,18 +32,7 @@ SECRET_KEY = "dqji)!xte^trgai!3c)_4)ftaoevwvbog-i&nl$#ef9xb+y*ab"
 DEBUG = True
 ENV = 'dev'
 
-ALLOWED_HOSTS = [
-    # QA
-    '.molo.unicore.io',
-    '.seed.p16n.org',
-    # Production
-    '.heyspringster.com',
-    'ninyampinga.com',
-    'www.ninyampinga.com',
-    'cewekeren.com',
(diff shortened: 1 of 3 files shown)


sewagodimo:
Looks good and ready to merge

gem/settings/base.py
```diff
@@ -32,7 +32,17 @@
...
+    # QA
+    '.molo.unicore.io',
+    '.seed.p16n.org',
+    # Production
+    '.heyspringster.com',
+    'ninyampinga.com',
+    'www.ninyampinga.com',
+    'cewekeren.com',
+    'www.cewekeren.com',
+]
```
nathanbegbie: @alexmuller is this something worth pulling in from environment variables instead of hard-coding it?
If we're adding URLs in our deployment tool (Mission Control or Spinnaker) would it make sense to have to add it there as well, instead of updating the code itself?

alexmuller: @nathanbegbie Yeah that's right, it's possible to do it that way. That would require much more of a deployment overhead but I can update to do that.

sewagodimo:
@nathanbegbie @alexmuller Thank you guys, Updated to use the environment var

PR #667, opened 2018-03-13
updated molo.commenting to version 6.2.8

sewagodimo:
updated requirements.txt to use the most recent version of molo.commmenting, 6.2.8

### DIFF
--- a/gem/tests/test_views.py
+++ b/gem/tests/test_views.py
@@ -337,6 +337,15 @@ class CommentingTestCase(TestCase, MoloTestCaseMixin):
                                        title='article 1',
                                        subtitle='article 1 subtitle',
                                        slug='article-1')
+        self.content_type = ContentType.objects.get_for_model(self.user)
+        self.french = SiteLanguageRelation.objects.create(
+            language_setting=Languages.for_site(self.site),
+            locale='fr',
+            is_active=True)
+        self.translated_article = self.mk_article_translation(
+            self.article, self.french,
+            title=self.article.title + ' in french',
+            subtitle=self.article.subtitle + ' in french')
 
     def create_comment(self, article, comment, user, parent=None):
         return MoloComment.objects.create(
@@ -368,6 +377,24 @@ class CommentingTestCase(TestCase, MoloTestCaseMixin):
         response = self.client.get('/sections-main-1/your-mind/article-1/')
         self.assertContains(response, "this is my alias")
         self.assertNotContains(response, "tester")
+
+    def test_anonymous_comment_translation(self):
+        self.user.profile.alias = 'this is my alias'
+        MoloComment.objects.create(
+            content_object=self.translated_article,
+            object_pk=self.translated_article.id,
+            content_type=ContentType.objects.get_for_model(
+                self.translated_article),
+            site=Site.objects.get_current(), user=self.user,
+            comment='This is another comment for French',
+            submit_date=timezone.now())
+        response = self.client.get('/locale/fr/')
+        response = self.client.get(
+            reverse('molo.commenting:more-comments', args=(
+                self.translated_article.pk,)))
+        self.assertContains(response, "This is another comment for French")
+        self.assertContains(response, "Anonyme")
+        self.assertNotContains(response, "this is my alias")
 
     def test_comment_distinguishes_moderator_user(self):
         self.user = User.objects.create_user(
--- a/gem/tests/test_views.py
+++ b/gem/tests/test_views.py
@@ -337,15 +337,6 @@ class CommentingTestCase(TestCase, MoloTestCaseMixin):
                                        title='article 1',
                                        subtitle='article 1 subtitle',
                                        slug='article-1')
-        self.content_type = ContentType.objects.get_for_model(self.user)
-        self.french = SiteLanguageRelation.objects.create(
-            language_setting=Languages.for_site(self.site),
-            locale='fr',
-            is_active=True)
-        self.translated_article = self.mk_article_translation(
-            self.article, self.french,
-            title=self.article.title + ' in french',
-            subtitle=self.article.subtitle + ' in french')
 
     def create_comment(self, article, comment, user, parent=None):
         return MoloComment.objects.create(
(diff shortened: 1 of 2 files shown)


codiebeulaine:
@sewagodimo can we include the test from the previous PR that checks that the alias doesn't show if commenting anonymously in a different language

Mitso:
🍰 @sewagodimo

gem/tests/test_views.py
```diff
@@ -369,6 +378,23 @@ def test_comment_shows_user_display_name(self):
...
+                self.translated_article),
+            site=Site.objects.get_current(), user=self.user,
+            comment='This is another comment for French',
+            submit_date=timezone.now())
+        response = self.client.get('/locale/fr/')
+        response = self.client.get(
+            reverse('molo.commenting:more-comments', args=(
+                self.translated_article.pk,)))
+        self.assertContains(response, "This is another comment for French")
+        # we test for translation of anonymous in project tests
```
codiebeulaine: this comment can be removed

sewagodimo:
@codiebeulaine I added the tests. So instead of checking that the translation is in french (because pootler complications) I just check that the word anonymous is not there, presumably it has been replaced with other text. Is this what you wanted

gem/tests/test_views.py
```diff
@@ -369,6 +378,24 @@ def test_comment_shows_user_display_name(self):
...
+                self.translated_article),
+            site=Site.objects.get_current(), user=self.user,
+            comment='This is another comment for French',
+            submit_date=timezone.now())
+        response = self.client.get('/locale/fr/')
+        response = self.client.get(
+            reverse('molo.commenting:more-comments', args=(
+                self.translated_article.pk,)))
+        self.assertContains(response, "This is another comment for French")
+        self.assertContains(response, "Anonyme")
```
saeedmb8: small typo here, I think you need to check for `Anonymous`

gem/tests/test_views.py
```diff
@@ -369,6 +378,24 @@ def test_comment_shows_user_display_name(self):
         self.assertContains(response, "this is my alias")
         self.assertNotContains(response, "tester")
 
+    def test_anonymous_comment_translation(self):
+        self.user.profile.alias = 'this is my alias'
```
saeedmb8: @sewagodimo If you are testing for Anonymous I don't think you need to add an alias to the user. However if you need to do that you have to save the user after assigning the Alias `self.user.profile.save()` but in this case you don't need `self.user.profile.alias = 'this is my alias'`

gem/tests/test_views.py
sewagodimo: this is the french article, it should be "Anonyme"

gem/tests/test_views.py
```diff
@@ -369,6 +378,37 @@ def test_comment_shows_user_display_name(self):
...
+            'submit_anonymously': '1',
+        })
+        self.client.post(
+            reverse('molo.commenting:molo-comments-post'), data)
+        [comment] = MoloComment.objects.filter(user=self.user)
+        print(comment)
+        self.assertEqual(comment.comment, 'Foo')
+        self.assertEqual(comment.user_name, 'Anonymous')
+
+    def test_anonymous_comment_translation(self):
```
saeedmb8: Since we are not testing for the poodle translation I don't think this test is necessarily

gem/tests/test_views.py
```diff
@@ -337,6 +337,15 @@ def setUp(self):
...
                                        slug='article-1')
+        self.content_type = ContentType.objects.get_for_model(self.user)
+        self.french = SiteLanguageRelation.objects.create(
+            language_setting=Languages.for_site(self.site),
+            locale='fr',
+            is_active=True)
+        self.translated_article = self.mk_article_translation(
+            self.article, self.french,
+            title=self.article.title + ' in french',
+            subtitle=self.article.subtitle + ' in french')
```
saeedmb8: I think we can remove these as well

saeedmb8:
🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰 🍰

PR #668, opened 2018-03-14
Surveys UI update

Mitso:


Mitso:

screen shot 2018-03-15 at 10 51 24
screen shot 2018-03-15 at 10 51 56

PR #669, opened 2018-03-14
Improve removal of email, location and mobile number

alexmuller:
1. Ensure they're never required
2. I found them in some more templates


sewagodimo:
I think this is good to go

PR #671, opened 2018-03-15
SAS 115 surveys ui default icon

Mitso:


Mitso:

screen shot 2018-03-15 at 16 13 50

PR #674, opened 2018-03-16
Add BBM redirect view and use it to detect BBM

alexmuller:
This is easier to maintain than separate domain names for BBM. My hope is that we can direct users to the `/bbm/` endpoint which will redirect them to the homepage but mark them as BBM users.

### DIFF
--- a/gem/context_processors.py
+++ b/gem/context_processors.py
@@ -2,8 +2,24 @@ from django.conf import settings
 
 
 def detect_bbm(request):
+    '''
+    Detect whether a request has come from BBM. BBM directs users to a /bbm/
+    endpoint which sets a cookie that we can check. Previously we did this
+    using a bbm subdomain, but that is complex to maintain.
+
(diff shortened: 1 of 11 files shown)


alexmuller:
Updated to address one of Codie's comments on Slack that the middleware needs to be updated to send to Analytics based on the cookie, I think this is ready for review on Monday. See you then!

nathanbegbie:
@alexmuller just had a thought - users need to be able to share articles via bbm, so there are share buttons for both the home page and article-specific shares.
In this case, we'll need to update `molo-gem/gem/templates/springster/patterns/basics/social-media/social-media_bbm.html` so that we include the redirect link in the shared content and the redirect will need to handle both home page and article page redirects.

alexmuller:
Gaaaah ✊ 

But yeah thanks, good point

alexmuller:
Updated to use `/bbm` as a prefix for redirecting. Took a little bit longer than I would have liked because it has some interesting repercussions for not being an open redirect that other people can abuse.

PR #676, opened 2018-03-19
QA feedback updates

Mitso:


Mitso:

Base - Added char counter on textfields.
Springster - Updated surveys default background

PR #680, opened 2018-03-22
Font-Face and Base64 Data-URI

Mitso:


Mitso:
@projectmushroom please review. 
There is currently FOUC and FOUT however it's very small. There is part 2 to fix this. 
 - Looking to test the difference in render blocking Font currently.

projectmushroom:
mmm I don't know about this one, the encoded inline fonts are heavy, the complied css is 1.2mb, the Baton fonts with their respected .eot .woff .woff2, all 9 of them are 594KB combined, and the browser won't even load all of them, just the first one supported, we can also probably get way with just using .woff and woff2, its widely supported nowadays and out feature-phones version uses the fallback anyway

PR #683, opened 2018-03-26
Updated springster templates to include anonymous alias templates

sewagodimo:


gem/templates/springster/comments/comment.html
```diff
@@ -11,7 +11,7 @@ <h5 class="comments__alias comments__alias-author">
         - {% trans "Moderator" %}
       </span>
   {% elif node.user_name.lower == 'anonymous' %}
-    {{node.user_name}}
```
OMosimege: Can I get a bit clarity on what the node.user used to do, and why its being changed to %trans "Anonymous"%, especially if its already called in line 17. Doesn't it make it redundant?

sewagodimo: It gets the user's alias but if the alais is anonymous it means the user chose to comment anonymously. If we use node.user_name it will not be translated for the other lanaguages. Trans has issues translating other values so that is why we just replace it with actual text and not just say  `trans node.user_name`

OMosimege: Looks great! +1

PR #684, opened 2018-03-27
SAS 118: Survey redesign updates

Mitso:


Mitso:

Make survey background color stronger on homepage vs survey page.

screen shot 2018-03-27 at 13 50 25

screen shot 2018-03-27 at 13 50 00

PR #687, opened 2018-03-29
Update fed readme status on pootle translation

Mitso:


gem/management/commands/molo_makemessages.py
```diff
@@ -0,0 +1,19 @@
+from __future__ import absolute_import, unicode_literals
```
saeedmb8: @Mitso lets not commit this now since it's not a generic script, we need to refactor this a bit so we would be able to give it different values for `fromDirectory`, `locale` and `ignore`

locale/en/LC_MESSAGES/django.po
```diff
@@ -2,7 +2,7 @@ msgid ""
 msgstr ""
```
saeedmb8: @Mitso can you re generate this file again with ignoring the `rules.py` as well, we don't need to add those since we don't translate the sentences that are shown in wagtail

locale/en/LC_MESSAGES/django.po
```diff
@@ -1722,54 +2065,133 @@ msgid ""
 "instructions for setting a new one."
 msgstr ""
 
+#: temp/temp_ve/core/templates/versions.html:8
```
saeedmb8: lets add the `versions.html` to the ignore list as well

locale/en/LC_MESSAGES/django.po
```diff
@@ -1681,6 +1989,19 @@ msgstr ""
 msgid "More about"
 msgstr ""
 
+#: temp/temp_ve/core/templates/modeladmin/includes/filter.html:2
```
saeedmb8: lets add `modeladmin/*` to ignore list as well

.gitignore
```diff
@@ -61,3 +61,4 @@ gem_test.db
 gem_test.db
 node_modules/
 db.sqlite3-journal
+gem/management/commands/molo_makemessages.py
```
saeedmb8: we don't need this here

Mitso:
@saeedmb8 updated, thank you!

PR #690, opened 2018-04-10
Add MAINTENANCE_MODE setting

alexmuller:
This overrides all the urlpatterns to serve a 503 error for every page with a simple template.

I think we should probably just add the different translations in a static way in this template.


gem/settings/base.py
```diff
@@ -33,6 +33,8 @@
 DEBUG = True
 ENV = 'dev'
 
+MAINTENANCE_MODE = environ.get('MAINTENANCE_MODE', '') == 'true'
```
codiebeulaine: what does the == true but do?

codiebeulaine:
@alexmuller can we write a test for this?

alexmuller:
@codiebeulaine I've tried to write a test for this but it's not possible to override the `MAINTENANCE_MODE` setting in the test because the urlconf is loaded when the app starts. I also tried to write a test for the view class directly but haven't figured out how to get that working yet.

codiebeulaine:
👍

PR #691, opened 2018-04-12
GEMMOLO 900: Main template render

Mitso:
Removed template blocks which uses Template inheritance collapsed template modules.
Also added


sewagodimo:
hey, may you please add a description for this PR

Mitso:
@sewagodimo 

Removed template blocks which uses Template inheritance collapsed template modules.
Also added

gem/settings/base.py
```diff
@@ -32,6 +32,7 @@
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = True
 ENV = 'dev'
+INTERNAL_IPS = ('127.0.0.1')
```
projectmushroom: if we're not adding the debug toolbar, should this still be here?

PR #694, opened 2018-04-16
Main and buttons templates collapse

Mitso:


Mitso:
@projectmushroom @sewagodimo 

Minimise and collapse templates for video contents.

gem/templates/springster/core/tags/media_listing_homepage.html
```diff
@@ -1,34 +1,3 @@
...
-          {% include "patterns/basics/buttons/sp_variations/button.html" with type="yellow" hyperlink=media.file.url text=buttontext attribute="download" tracktitle=media.title tracktype="Audio" %}
-        </div>
-      </div>
-    {% endif %}
-  {% endif %}
-{% endfor%}
-{% endblock %}
-  
\ No newline at end of file
+  {% include "patterns/components/media/media-player.html" %}
```
sewagodimo: I thought we were reducing the number of include statements

gem/templates/springster/core/main.html
```diff
@@ -30,7 +31,22 @@
     {% include 'core/tags/latest_listing_homepage.html' with articles=latest_articles|slice:"0:3" %}
     {% include "patterns/components/article-teasers/sp_variations/carousel-teaser-listing.html" with tag=tag_nav_data.tags_list.0.0 tag_articles=tag_nav_data.tags_list.0.1 %}
 
-    {% media_listing_homepage %}
+    {% if is_via_freebasics %}
```
sewagodimo: so this code was not working at all. The list of media files cannot be accessed from he main.html file. They are accessible from the media_listing_homepage.html and it already has code for dealing with freebasics sites

PR #695, opened 2018-04-16
Add health URL when in maintenance mode

alexmuller:
Mesos/Marathon still needs to be able to check the health of the app.


codiebeulaine:
👍

PR #699, opened 2018-04-19
collapse comment templates

projectmushroom:


gem/templates/springster/comments/reply.html
```diff
@@ -1,31 +1,208 @@
...
-    {% endfor %}
-  {% include "patterns/basics/pagination/pagination-comments.html"%}
+  </div>
   <div id="comment-block">
     <div class="heading heading__component">
       <h2>{% trans "Add your reply" %}</h2>
     </div>
     {% if request.user.is_authenticated %}
-      {% include "comments/form.html" with node=comment %}
+      <form id="comment-form" action="{% url 'molo.commenting:molo-comments-post' %}#comments-list" method="post" class="comments__form" novalidate>
```
Mitso: @projectmushroom with the forms I do think we can reuse although we can reuse plenty of things and paying for it. Thoughts?

Mitso:
🍰 🍰 @projectmushroom

gem/templates/springster/comments/reply.html
projectmushroom: mmm, I think its best if we try to collapse everything, If we don't, like you said, we can reuse plenty of things, and yes pay for it in the end.  😞

PR #707, opened 2018-04-24
Update broken link image

Mitso:


Mitso:

Section image shows broken image if it is not uploaded.
Fix: check if section image exist first.

PR #709, opened 2018-04-24
collapse profile templates

projectmushroom:


gem/templates/springster/profiles/editprofile.html
```diff
@@ -1,8 +1,18 @@
 {% extends "base.html" %}
 {% block content %}
 <div class="profiles">
-    {% trans "Edit your Profile" as headingtext %}
+
```
Mitso: Does this needs to be collapsed?

projectmushroom: oh snap, updating....

PR #712, opened 2018-05-01
Fix silent template exceptions

alexmuller:
##  Override main page meta tags to remove silent exceptions

When I load the homepage in development I get thousands of log lines of exceptions and 264 instances of the word "Traceback".

This is because we're accessing lots of properties in the meta tags which the Main model doesn't have.

To fix it I've overriden the meta tags on the main page so that we don't try and access those properties, which gets rid of the exceptions. Locally it takes the number of tracebacks from 264 to 132.

##  Override search input on search results page

Currently we're trying to access `search_query` in the base template which causes a silent exception on every page except the search results page.

Instead we should override that bit of the template on the search results page to maintain the behaviour and prevent the exceptions.


codiebeulaine:
👍

PR #719, opened 2018-05-04
Update survey templates

Mitso:


Mitso:
@projectmushroom a review please.
Truncate long sections as a result of template collapsing.

gem/templates/springster/surveys/molo_survey_page.html
```diff
@@ -2,545 +2,223 @@
...
-
-  {# include template "patterns/basics/surveys/surveys-modules/surveys_gcn-heading.html" #}
-    {% with pagetype=self %}
-      {% if pagetype.image %}
-          {% image pagetype.image width-98 as tmp_photo_small %}
-          {% image pagetype.image width-320 as tmp_photo_large %}
-          <div class="gcn-competition__image">
-              <img src="{{ tmp_photo_small.url }}" srcset="{{ tmp_photo_large.url }}" alt="{{ survey.image.title }}" class="gcn-competition__dynamic" />
-          </div>
+  <!--HEADING-->
```
projectmushroom: We're adding more and more comments in our templates, might be a good idea to use django comments `{# comment won't be rendered #}` instead of HTML comments that are viewable on page source

PR #720, opened 2018-05-07
Removing join banner duplications

sewagodimo:


Mitso:
@sewagodimo look good to me

🍰 🍰 🍰 🍰 🍰 🍰

PR #725, opened 2018-05-09
Surveys terms page

Mitso:


Mitso:

screen shot 2018-05-09 at 15 35 30

gem/styles/gem-springster/modules/surveys/_m-surveys.scss
```diff
@@ -49,6 +48,7 @@
           }
       }
       &__page {
+        //Where is this located?
```
projectmushroom: note to self? 😄

PR #747, opened 2018-05-21
Reverse VIEW_PROFILE_URL and EDIT_PROFILE_URL in settings

alexmuller:
These are sometimes set to external URLs (when using OIDC to log in). We can't pass external URLs into the url template tag so we need to make them consistent and just render the value in the template.


codiebeulaine:
WTIH

PR #748, opened 2018-05-22
Store users auth service uuid on profile

KaitCrawford:
In order to link an existing Springster profile with a user that has just logged in via the Auth Service we store the uuid for their auth service account on their profile on first login.
On later logins we then just use that field to match the user profile.

This is dependent on #742


gem/backends.py
```diff
@@ -33,6 +33,11 @@ def _update_user_from_claims(user, claims):
     user.email = claims.get("email", "")
     user.save()
 
+    # Ensure the profile is linked to their auth service account using the uuid
+    if user.profile.auth_service_uuid is None:
```
codiebeulaine: what if the user does not have a profile? This will return an error

gem/backends.py
```diff
@@ -33,6 +33,11 @@ def _update_user_from_claims(user, claims):
...
+        user.profile.auth_service_uuid = claims["sub"]
```
codiebeulaine: can we user claims.get("sub"), just to keep with the styling of this backend

gem/backends.py
```diff
@@ -62,14 +67,29 @@ def filter_users_by_claims(self, claims):
         """
         uuid = claims["sub"]
         try:
-            kwargs = {USERNAME_FIELD: uuid}
+            kwargs = {'profile__auth_service_uuid': uuid}
```
codiebeulaine: can we put an except here for if the profile does not exist, and things for what do if that happens?

codiebeulaine:
👍 wtih

PR #749, opened 2018-05-22
remove --nomigrations from setup, add gem test base

codiebeulaine:


nathanbegbie:
@codiebeulaine what does removing `--nomigrations` do for the tests?

codiebeulaine:
@nathanbegbie it runs the test without migrations.

PR #757, opened 2018-05-24
Added logut settings for OIDC_LOGOUT_URL

sewagodimo:
ogout settings for OIDC_LOGOUT_URL


gem/settings/base.py
```diff
@@ -57,6 +57,7 @@
 LOGOUT_REDIRECT_URL = environ.get('LOGOUT_REDIRECT_URL')
 WAGTAIL_REDIRECT_URL = environ.get('WAGTAIL_REDIRECT_URL', '')
 OIDC_OP_LOGOUT_URL_METHOD = "gem.utils.provider_logout_url"
+OIDC_OP_LOGOUT_URL = "http://authentication-service.qa-hub.ie.gehosting.org/openid/end-session/"
```
codiebeulaine: @sewagodimo this should get the variable from the environment like the rest of the settings. And then this value can be put in the environment through spinnaker

codiebeulaine:
WTIH 🍰

PR #765, opened 2018-06-01
GEMMOLO-1028 implement auth service roles

saeedmb8:


gem/backends.py
```diff
@@ -56,18 +56,30 @@ def _update_user_from_claims(user, claims):
...
+    auth_service_roles = set(claims.get("roles", []))
+    wagtail_groups = set(group.name for group in user.groups.all())
+
+    # If the user has any role, add the wagtail group equivalent
+    # to that role to the user
+    if auth_service_roles:
+        groups_to_add = auth_service_roles - wagtail_groups
+        groups_to_remove = wagtail_groups - auth_service_roles
+        for group_name in groups_to_add:
+            if group_name == 'Product Tech Admin':
```
jpaidoussi: Perhaps "Product Tech Admin" should be defined as a constant somewhere rather than just as a raw value in the middle of a function scope.

gem/migrations/0030_create_groups_and_permissions.py
```diff
@@ -0,0 +1,191 @@
...
+    # remove the existing groups
+    Group.objects.all().delete()
+
+    # **** Get IndexPages ****
+    sections = SectionIndexPage.objects.all()
+    banners = BannerIndexPage.objects.all()
+    footers = FooterIndexPage.objects.all()
+    polls = PollsIndexPage.objects.all()
+    surveys = SurveysIndexPage.objects.all()
+    yourwords = YourWordsCompetitionIndexPage.objects.all()
```
jpaidoussi: How big do you expect some of these object lists to be? Could it become a memory problem to load them all into memory upfront?

saeedmb8: @jpaidoussi the largest multi site cluster has 30 sites, so for that app it would be a list of 30 objects for each of the Index Pages, I don't expect it to be a problem, what do you thing?

jpaidoussi: I don't think that would cause a problem, but it may help to use `iterator()` anyway: https://docs.djangoproject.com/en/2.0/ref/models/querysets/#iterator

saeedmb8:
@jpaidoussi ready for re-review

gem/backends.py
```diff
@@ -56,18 +56,30 @@ def _update_user_from_claims(user, claims):
...
+                user.is_staff = True
+                user.is_superuser = True
+                user.save()
+            else:
+                try:
+                    user.groups.add(Group.objects.get(name=group_name))
+                except Group.DoesNotExist:
+                    LOGGER.debug("Group {} does not exist".format(group_name))
+        # Remove the user's revoked role
+        for group_name in groups_to_remove:
```
KaitCrawford: Since we treat the Superuser group differently above shouldn't we treat it differently here too?
If a user is removed from the superuser group is there any reason not to set their is_superuser and is_staff flags to false?

saeedmb8: @KaitCrawford good catch 👌

PR #770, opened 2018-06-06
added fixes for the login, url

sewagodimo:
The login button should take users to a login page with a springster theme


gem/settings/base.py
```diff
@@ -61,7 +61,9 @@
 OIDC_OP_LOGOUT_URL = environ.get("OIDC_OP_LOGOUT_URL", "")
 
 if USE_OIDC_AUTHENTICATION:
-    LOGIN_URL = 'oidc_authentication_init'
+    LOGIN_URL = (
+        "%s/login/?theme=%s&hide=end-user&redirect_url=%s" % (
```
codiebeulaine: instead of defining the url like this, can we do a reverse of 'oidc_authentication_init' which should give us the url path it points to, and then add the theme and redirect uri to that? This safeguards us against mozilla oidc changing the url path

sewagodimo: Thanks

PR #773, opened 2018-06-07
Update auth links with correct settings for multisite

codiebeulaine:
We need to update the auth service links with the correct parameters dynamically based on the `request.site`.

I'm not sure if this is the best way to do it, we could keep these links in the settings file and use middleware to update it as an alternative? but these values are only called in the templates and bad practice to use middleware to update settings?

Any thoughts?

### DIFF
--- a/gem/context_processors.py
+++ b/gem/context_processors.py
@@ -33,12 +33,39 @@ def detect_freebasics(request):
 
 
 def compress_settings(request):
+    REGISTRATION_URL = settings.REGISTRATION_URL
+    EDIT_PROFILE_URL = settings.EDIT_PROFILE_URL
+    VIEW_PROFILE_URL = settings.VIEW_PROFILE_URL
+
+    if settings.USE_OIDC_AUTHENTICATION:
+        site = request.site
+        if not hasattr(site, "oidcsettings"):
+            raise RuntimeError(
+                "Site {} has no settings configured.".format(site))
+
+        oidc_settings = site.oidcsettings
+        if oidc_settings:
+            REGISTRATION_URL = (
+                "%s/registration/?theme=%s&hide=end-user&redirect_uri=%s" % (
+                    settings.OIDC_OP, settings.THEME,
+                    oidc_settings.wagtail_redirect_url))
+            VIEW_PROFILE_URL = (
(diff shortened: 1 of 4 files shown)


gem/context_processors.py
```diff
@@ -33,12 +33,39 @@ def detect_freebasics(request):
...
+            VIEW_PROFILE_URL = (
+                "%s/profile/edit/?theme=%s&redirect_uri=%s&client_id=%s" % (
+                    settings.OIDC_OP, settings.THEME,
+                    oidc_settings.wagtail_redirect_url,
+                    oidc_settings.oidc_rp_client_id))
+            EDIT_PROFILE_URL = (
+                "%s/profile/edit/?theme=%s&redirect_uri=%s&client_id=%s" % (
+                    settings.OIDC_OP, settings.THEME,
+                    oidc_settings.wagtail_redirect_url,
+                    oidc_settings.oidc_rp_client_id))
```
sewagodimo: what about log out? 
and login?
Are we using the oidc urls for that?

gem/settings/base.py
```diff
@@ -63,15 +63,6 @@
 if USE_OIDC_AUTHENTICATION:
```
sewagodimo: what about passing the theme to the login url or should we do that in another PR??

gem/context_processors.py
```diff
@@ -33,12 +33,39 @@ def detect_freebasics(request):
...
+        if not hasattr(site, "oidcsettings"):
+            raise RuntimeError(
+                "Site {} has no settings configured.".format(site))
+
+        oidc_settings = site.oidcsettings
+        if oidc_settings:
+            REGISTRATION_URL = (
+                "%s/registration/?theme=%s&hide=end-user&redirect_uri=%s" % (
+                    settings.OIDC_OP, settings.THEME,
+                    oidc_settings.wagtail_redirect_url))
```
sewagodimo: should registration not be redirecting to the home page?

codiebeulaine: @sewagodimo  the wagtail redirect url can be set to the homepage in the OIDC settings

gem/settings/base.py
codiebeulaine: Did that in a separate PR, it is handled differently to everything else

PR #778, opened 2018-06-11
GER redesign setup

Mitso:


Mitso:
@sewagodimo please review

gem/templates/rwanda/core/footer_page.html
```diff
@@ -0,0 +1,36 @@
...
+<div class="articles nav">
+  <h3 class="article-title">{{self.title}}</h3>
+  <p>{{self.subtitle}}</p>
+
+  {% if self.image %}
+  <div class="banner">{% image self.image width-320 %}</div>
+  {% endif %}
+
+<article>
+  {% for block in self.body %}
```
sewagodimo: this this a django template block instance? `{%block content %} {%endblock%}`
I.E. are we getting all the blocks in the template body and rendering them?
If not the word choice is a bit confusing

gem/templates/rwanda/core/footer_page.html
```diff
@@ -0,0 +1,36 @@
...
+      {% if block.block_type == 'heading' %}
+          <h1>{{ block.value }}</h1>
+      {% else %}
+      {% if block.block_type == 'image' %}
+        {% image block.value width-320 %}
+      {% else %}
+      {% if block.block_type == 'numbered_list' %}
+        <ol>
+        {% for item in block.value %}
```
sewagodimo: item could just be named value

gem/templates/rwanda/core/footer_page.html
Mitso: @sewagodimo is currently a clone of the base template (initial setup) - will be fixed on subsequent tickets

gem/templates/rwanda/core/footer_page.html
Mitso: Same as above comment

gem/templates/rwanda/core/footer_page.html
Mitso: Updated

PR #779, opened 2018-06-11
Mark user as staff when they are added to a group and add yourwords entry permission

saeedmb8:


gem/backends.py
```diff
@@ -73,19 +73,25 @@ def _update_user_from_claims(user, claims):
                 try:
                     wagtail_group = Group.objects.get(name=group_name)
                     user.groups.add(wagtail_group)
+                    user.is_staff = True
+                    user.save()
```
KaitCrawford: I think this needs to be in an if block. Otherwise you're setting it even if they are already staff and that seems like unnecessary db writes

gem/migrations/0031_add_view_yourwords_entry_to_groups.py
```diff
@@ -0,0 +1,35 @@
...
+    Group = apps.get_model('auth.Group')
+    Permission = apps.get_model('auth.Permission')
+
+    view_yourwords_entry = Permission.objects.get(codename='can_view_entry')
+    group_names = ('product_admin', 'data_admin', 'data_viewer',
+                   'content_admin', 'content_admin')
+    for group_name in group_names:
+        group = Group.objects.get(name=group_name).first()
+        group.permissions.add(view_yourwords_entry)
+
```
KaitCrawford: I'm confused by this method. How does we know that it the `can_view_entry` permission only exists for the YourWords model?

saeedmb8: @KaitCrawford you are right, not that I'm thinking about it, it is confusing, so the full permission is `yourwords | YourWords Competition Entry | Can View Entry` but I think I should have added the code name as `can_view_yourwords_entry`.

PR #780, opened 2018-06-12
Uupdate profiles form fields settings

Mitso:


Mitso:
@sewagodimo @saeedmb8  a review please.

Update templates with settings enabled fields.

gem/templates/base/profiles/viewprofile.html
```diff
@@ -34,11 +34,25 @@ <h3>{% trans "Your Profile" %}</h3>
...
+        {% endif %}
+
+        {% if settings.profiles.UserProfilesSettings.show_email_field %}
+          <p>{% trans "Email Address:" %}
+          <span>{{ user.email | default:"Not set" }}</span></p>
+        {% endif %}
+
+        {% if settings.profiles.UserProfilesSettings.activate_location %}
+            <p>{% trans "Location:" %}
+            <span>{{ user.profile.location | default:"Not set" }}</span></p>
```
sewagodimo: will the default be translated?

gem/templates/base/profiles/register.html
```diff
@@ -63,6 +63,24 @@ <h3>{% trans "Sign up" %}</h3>
...
+            <p class="helptext">{% trans "(e.g. 0900000013)" %}</p>
+          </fieldset>
+        {% endif %}
+
+        {% if settings.profiles.UserProfilesSettings.show_email_field %}
+          <fieldset>
+            <label for="id_email">{% trans "Enter your email address" %}</label>
+            {{ form.email.errors}}
+            {{ form.email }}
+            <p class="helptext">{% trans "(e.g. example@foo.com)" %}</p>
```
sewagodimo: why don't we set set the help test in the field model?

Mitso: We need to select and style help-text separately here.

PR #781, opened 2018-06-12
added the alias check for users

sewagodimo:


gem/backends.py
```diff
@@ -41,6 +41,13 @@ def _update_user_from_claims(user, claims):
         user.profile = UserProfile(user=user)
         user.profile.site = Site.objects.get(is_default_site=True)
 
+    # check if the user has an alias
+    # if so. ensure that the alias is eaul to the user_name
+    if user.profile.alias is None or user.profile.alias == "":
+        user.profile.alias = user.first_name
```
KaitCrawford: I think we should probably be using user.username here rather than user.first_name. Otherwise we'd be displaying a users firstname on their comments.

gem/backends.py
```diff
@@ -41,6 +41,13 @@ def _update_user_from_claims(user, claims):
...
+    elif user.profile.alias != user.first_name:
+        user.profile.alias = user.first_name
+
```
KaitCrawford: I think the section above would fit better in the block that has the comment `Synchronise a user's profile data`
Just so we keep everything close together.

gem/tests/test_auth.py
```diff
@@ -101,6 +102,7 @@ def test_add_superuser_role_from_claims(self):
                           'e2556752-16d0-445a-8850-f190e860dea4')
         self.assertEquals(user.profile.gender, 'f')
         self.assertEquals(user.profile.date_of_birth, date(1988, 5, 22))
+        self.assertEquals(user.profile.alias, 'testgivenname')
```
KaitCrawford: Please also add a check to confirm the alias is set even if it was previously None.

gem/tests/test_auth.py
```diff
@@ -118,6 +118,25 @@ def test_add_user_role_from_claims(self):
         self.assertEquals(user.groups.all().count(), 1)
         self.assertEquals(user.groups.first().name, 'data_admin')
 
+    def test_display_username_as_alias_from_claims(self):
```
KaitCrawford: Just a tip for the test name: make sure it's accurate for the test. In this case you're not testing the display or updating the alias from the claims. You're testing that the alias is always set to be the same as the username.

gem/tests/test_auth.py
```diff
@@ -118,6 +118,25 @@ def test_add_user_role_from_claims(self):
...
+        roles = ['data_admin', ]
+        claims = {
+            'roles': roles,
+            'given_name': 'testgivenname',
+            'family_name': 'testfamilyname',
+            'email': 'test@email.com',
+            'sub': 'e2556752-16d0-445a-8850-f190e860dea4'}
+        user = get_user_model().objects.create(
+            username='testuser', password='password')
+
```
KaitCrawford: Please add an assert here to show that the alias is empty/None at this point.
That way the test is 100% explicit about which code blocks it's testing

PR #782, opened 2018-06-12
add auto login after rego

codiebeulaine:


gem/context_processors.py
```diff
@@ -49,7 +50,8 @@ def compress_settings(request):
                 "%s/registration/?theme=%s&hide=end-user&redirect_uri=%s"
                 "&client_id=%s" % (
                     settings.OIDC_OP, settings.THEME,
-                    oidc_settings.wagtail_redirect_url,
+                    request.site.root_url
+                    + reverse('oidc_authentication_init'),
```
saeedmb8: @codiebeulaine should we maybe use urlparse.urljoin(url1, url2) here instead

saeedmb8:
small suggestion other wise looks good 👍

PR #783, opened 2018-06-14
Add language code to requests to the auth service

KaitCrawford:
In order to translate pages on the auth service to the user's preference we need to send their preference through as part of the url.
This adds it for the Login, profile edit, profile view and registration requests

### DIFF
--- a/gem/context_processors.py
+++ b/gem/context_processors.py
@@ -40,6 +40,7 @@ def compress_settings(request):
 
     if settings.USE_OIDC_AUTHENTICATION:
         site = request.site
+        language = getattr(request, 'LANGUAGE_CODE', 'en')
         if not hasattr(site, "oidcsettings"):
             raise RuntimeError(
                 "Site {} has no settings configured.".format(site))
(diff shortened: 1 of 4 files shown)


gem/context_processors.py
```diff
@@ -40,6 +40,7 @@ def compress_settings(request):
 
     if settings.USE_OIDC_AUTHENTICATION:
         site = request.site
+        language = getattr(request, 'LANGUAGE_CODE', 'en')
```
sewagodimo: Do we want it to always be english or should we not set a value in settings, like the theme

gem/context_processors.py
```diff
@@ -48,21 +49,23 @@ def compress_settings(request):
...
-                    oidc_settings.oidc_rp_client_id))
+                    oidc_settings.oidc_rp_client_id, language))
             EDIT_PROFILE_URL = (
-                "%s/profile/edit/?theme=%s&redirect_uri=%s&client_id=%s" % (
+                "%s/profile/edit/?theme=%s&redirect_uri=%s&client_id=%s"
+                "&language=%s" % (
                     settings.OIDC_OP, settings.THEME,
                     oidc_settings.wagtail_redirect_url,
-                    oidc_settings.oidc_rp_client_id))
+                    oidc_settings.oidc_rp_client_id, language))
```
sewagodimo: what about login and log out?

KaitCrawford: Login is handled differently and the code added to views.py covers that.
The logout link doesn't take us to any screens on the auth service so I don't think it's necessary

gem/context_processors.py
KaitCrawford: Good point

gem/context_processors.py
sewagodimo: okay cool

KaitCrawford: @codiebeulaine can you confirm that my comment above about the logout link is correct?

PR #792, opened 2018-06-22
Section template redesign

Mitso:


Mitso:
@sewagodimo @KaitCrawford a review please.

Please pass the empty class placeholders at the moment.

PR #803, opened 2018-07-04
Banner notification for QA

Mitso:


Mitso:

gem/templates/rwanda/core/main.html
```diff
@@ -13,10 +13,10 @@
       </p>
     {% endif %}
 
-    {% if 'ninyampinga.com' in request.get_host %}
+    {% if 'ninyampinga.com' in request.get_host or 'ninyampinga-rwanda.qa-hub.ie.gehosting.org' in request.get_host %}
```
sewagodimo: do we want to leave this here of add it as a var setting?

Mitso: @sewagodimo Good point, updated to a template variable.

PR #809, opened 2018-07-05
Articles refinement

Mitso:


Mitso:
@sewagodimo @KaitCrawford a review please.

Update article teaser alignment as per client feedback.
Adjust comments layout
Update related, recommended and search results - article teasers

PR #810, opened 2018-07-06
Search default decoration 

Mitso:


Mitso:

Remove default search input decoration

PR #814, opened 2018-07-09
Section QA Feedback

Mitso:


Mitso:
@sewagodimo @KaitCrawford please review

Reduce spacings
Update article teaser character lengths

PR #815, opened 2018-07-09
Tags page update

Mitso:


Mitso:
@KaitCrawford @sewagodimo please review

screen shot 2018-07-10 at 10 05 11

PR #816, opened 2018-07-10
Input text wrap

Mitso:


Mitso:

screen shot 2018-07-10 at 11 00 07

PR #819, opened 2018-07-10
Section pagination + general 

Mitso:


Mitso:
@KaitCrawford @sewagodimo please review 

Updated search templates
Pagination on search and section pages

PR #821, opened 2018-07-11
Surveys QA UX - Feedback

Mitso:


Mitso:
@sewagodimo @KaitCrawford please review 

Updated the surveys page to show surveys when user is not looged in as per setting

gem/templates/rwanda/surveys/surveys_list.html
```diff
@@ -106,38 +106,45 @@ <h5 class="heading__subheading">
         </div>
         <div class="surveys__type--polls-separator"></div>
       {% else %}
-        {% if user.is_authenticated and user.is_active or request.is_preview or survey.allow_anonymous_submissions %}
+        {% if not user.is_authenticated and not request.is_preview and not survey.allow_anonymous_submissions and settings.gem.GemSettings.show_join_banner %}
+          {% comment %}
+            <div class="heading">
+              <h1 class="heading__component">
+                SETTING ABOVE| Do not show survey - just show join banner<h1>
```
sewagodimo: does this not need to be translated

Mitso: @sewagodimo No, this is just a comment for future reference to not duplicate code

PR #824, opened 2018-07-12
Login page and registration page fixes

KaitCrawford:
Added fields back to the registration page
Styled the login page and the registration 'done' page


KaitCrawford:
Login page
login_page
'Done' page
profile_done_page

KaitCrawford:
@Mitso I know we discussed this on the call but please advise:
Note the differences in the links at the bottom of the two pages.
I opted to differentiate between the nav links (not bold) and the call-to-action links (bold) but I'm not sure it's the right decision. I feel like the call-to-action links look better but the nav links are more consistent with other pages on the site.
Should I differentiate like this or should I pick of the two and stick with it?

KaitCrawford:
Changed buttons on the 'Done' page
nav_buttons

KaitCrawford:
I accidentally added the login_success template in one of the commits. Here is what it looks like:
login_success

gem/templates/rwanda/profiles/register.html
```diff
@@ -68,7 +113,7 @@ <h4 class="heading heading--subtitle">
     <fieldset>
       {{ form.terms_and_conditions }}
       {% if settings.profiles.UserProfilesSettings.terms_and_conditions %}
-      <a href="{% pageurl settings.profiles.UserProfilesSettings.terms_and_conditions %}" for="id_terms_and_conditions" class="profiles__terms">{% trans "I accept the Terms and Conditions" %}</a>
+      <label for="id_terms_and_conditions" class="profiles__terms"><a href="{% pageurl settings.profiles.UserProfilesSettings.terms_and_conditions %}">{% trans "I accept the Terms and Conditions" %}</a></label>
```
Mitso: @KaitCrawford I this clickable when Terms link is enabled?

gem/templates/rwanda/profiles/register.html
```diff
@@ -44,20 +44,65 @@ <h4 class="heading heading--subtitle">
...
+        <span class="input-hint">{% trans "Only you will see this." %}</span>
+      </fieldset>
+    {% endif %}
+
     {% if settings.profiles.UserProfilesSettings.show_security_question_fields  %}
       {% for field in form %}
         {% if field in form.security_questions %}
           <fieldset>
-            <label for="id_email">{% trans field.label_tag %}</label>
+            {% trans field.label_tag %}
```
Mitso: @KaitCrawford maybe add a label with the correct For input ID

PR #826, opened 2018-07-13
Style Rwanda forgot password page

KaitCrawford:


KaitCrawford:
Forgot password page:
forgot_password_page

gem/styles/gem-rwanda/modules/_m-profiles.scss
```diff
@@ -14,6 +14,11 @@
       margin-top: 0px;
       color: $color-pink;
     }
+    & .heading--subtitle {
```
Mitso: @KaitCrawford I think you can remove the &, it points to the parent selector if child selector is nested.

KaitCrawford: oh! I thought it was just a placeholder, string concatenation thing :joy:

PR #828, opened 2018-07-13
Style Rwanda profile view and edit pages

KaitCrawford:
This updates the style of the profile view and edit pages as well as the change password page.


KaitCrawford:
View Profile page:
profile_view

Edit profile page:
edit_profile_1
profile_edit_2

Change pin page:
change_pin_page

gem/templates/rwanda/profiles/viewprofile.html
```diff
@@ -0,0 +1,71 @@
...
+  <div class="profiles">
+    <div class="profiles__account"
+    <fieldset>
+      <label class="profiles__description--label">{% trans "When you type comments your name will appear as:" %}</label>
+      <p class="profiles__description profiles__alias">{{ user.profile.alias | default:" Anonymous " }}</p>
+    </fieldset>
+
+    {% trans "Not set." as defaultText %}
+    <fieldset>
+      <label class="profiles__description--label">{% trans "Date of Birth: " %}</label>
```
Mitso: @KaitCrawford is there a specific reason the classes to modify profiles__description has different name? They could just use the same modifier class

gem/styles/gem-rwanda/modules/_m-profiles.scss
```diff
@@ -27,23 +27,31 @@
     }
   }
   &__account {
-    padding: 0 $spacer-mid;
+    padding: $spacer-small $spacer-mid;
+    & .call-to-action{
```
Mitso: can remove the & here as well and just use .call-to-action

gem/templates/rwanda/profiles/viewprofile.html
KaitCrawford: @Mitso I'm afraid I don't understand. Which modifier class should they be using? (tbh I'm not sure what a 'modifier class' is)

Mitso:
@KaitCrawford looks good to me. 🍰 🍰 🍰

PR #829, opened 2018-07-16
Article tag alignment

Mitso:


Mitso:
@sewagodimo please review 

Update tags  to inline display

PR #831, opened 2018-07-16
General ui tweaks

Mitso:


Mitso:
@sewagodimo @KaitCrawford @moh-moola 
Please review. Adding GA script and small css tweaks

PR #833, opened 2018-07-16
Tags settings

Mitso:


Mitso:
@moh-moola please review 

Added tags settings.

gem/templates/rwanda/core/section_page.html
```diff
@@ -58,17 +58,20 @@ <h3 class="article-teaser__title">
...
-                {% for tag in tags %}
-                {% if not tag.title.20 %}
-                  <li class="teaser-stacklist-tags__item">
-                    <a href="{% url 'tags_list' tag.slug %}" class="teaser-stacklist-tags__anchor">
-                      {{ tag.title }}
-                    </a>
-                  </li>
-                  {% endif %}
-                {% endfor %}
+              {% if settings.core.SiteSettings.enable_tag_navigation %}
```
moh-moola: `{% if tags and settings.core.SiteSettings.enable_tag_navigation %}`
nothing major but the if block could be simplified

PR #834, opened 2018-07-17
Update tags article 

Mitso:


Mitso:
@moh-moola please review.

Remove container div when tag is not enabled.

PR #838, opened 2018-07-19
Performance and optimization validations

Mitso:


gem/styles/gem-rwanda/modules/_m-comments.scss
```diff
@@ -1,19 +1,22 @@
 .comments {
   padding: 0  $spacer-mid;
-  &:before {
+  /*&:before {
```
Mitso: To test this out before removal.

Mitso:
@sewagodimo @KaitCrawford @moh-moola 

Please review.
Markup validations update

PR #840, opened 2018-07-19
use env variable for GA log setting

KaitCrawford:


KaitCrawford:
Thanks @Mitso. Yeah, the second argument just supplies a default value in case the key we're looking for doesn't exist.

PR #846, opened 2018-07-24
Removing sensitive information from GA report

sewagodimo:


gem/settings/base.py
```diff
@@ -443,6 +443,9 @@
     'DEFAULT_FROM_EMAIL', 'support@moloproject.org')
 
 GOOGLE_ANALYTICS = {}
+
+SERVICE_DIRECTORY_API_BASE_URL = environ.get(
+    'SERVICE_DIRECTORY_API_BASE_URL', '')
```
KaitCrawford: @sewagodimo what is this for? where is it used?

sewagodimo: nice catch

PR #848, opened 2018-07-25
Translation audit

Mitso:


Mitso:
@codiebeulaine @sewagodimo @moh-moola please review 

Updated .po file translations

PR #849, opened 2018-07-25
General Updates

Mitso:


Mitso:
@sewagodimo @KaitCrawford @codiebeulaine  please review.

Updated .po file and general style.

PR #855, opened 2018-07-30
changed ga logging settings

sewagodimo:


KaitCrawford:
Remember after landing this you'll have to go to the pipeline in Spinnaker and change the setting on QA since this is checking for all lower case `true` and I think we put the setting in as `True`

PR #861, opened 2018-08-06
Media and Article update

Mitso:


Mitso:
@codiebeulaine @sewagodimo @moh-moola  please review.
1. Removed Article page heading and subtitle character limit.
2. Media alignment styles

PR #865, opened 2018-08-13
Update image size

Mitso:


Mitso:
@codiebeulaine @sewagodimo @moh-moola 
Please review

PR #868, opened 2018-08-14
Search input update

Mitso:


Mitso:
@sewagodimo please +1

PR #869, opened 2018-08-14
Media player in article

Mitso:


Mitso:
@codiebeulaine @sewagodimo @moh-moola please review 
Media update in article
screen shot 2018-08-14 at 13 01 29

PR #872, opened 2018-08-15
Template translation

Mitso:


Mitso:
@sewagodimo @codiebeulaine please review.

PR #879, opened 2018-08-20
Investigate search results counter

Mitso:


Mitso:
@sewagodimo @moh-moola please review

PR #885, opened 2018-08-23
add middleware to log header dump

codiebeulaine:


gem/middleware.py
```diff
@@ -35,6 +35,14 @@ def process_request(self, request):
             del request.META['HTTP_ACCEPT_LANGUAGE']
 
 
+class LogHeaderInformationMiddleware(object):
+
+    def process_request(self, request):
+        print ('---------- Header Dump -------------')
```
KaitCrawford: Are you sure print statements show up in the logs? I think we've had trouble with this before. It might be better to look at how Sewa did this for the GA URL stuff here https://github.com/praekelt/django-google-analytics/blob/develop/google_analytics/utils.py#L141-L142

codiebeulaine: changed

PR #923, opened 2018-10-17
Responsive header component

Mitso:


Mitso:
@moh-moola @KaitCrawford @sewagodimo Please review 

screen shot 2018-10-19 at 12 24 22

Mitso:
Rearrange header components and style as per the responsive designs.

PR #924, opened 2018-10-18
version pump molo.polls: 7.0.1

sewagodimo:


requirements.txt
```diff
@@ -20,6 +20,8 @@ django-modelcluster==3.1
 mozilla_django_oidc==1.0.0
 # Required by ImageHash, install a working version
 scipy==0.19.1
+# Required by Elastic-search and requests
```
KaitCrawford: I think the comment should be more like "greater versions break the requests dependency"

PR #927, opened 2018-10-19
Responsive Layout headerComponent - Search

Mitso:


Mitso:
@moh-moola @sewagodimo please review.
Updated the header search markup and styles.

PR #931, opened 2018-10-22
Respsonsive Layout - Banners

Mitso:


Mitso:
@moh-moola @KaitCrawford @sewagodimo 
Please review 

screen shot 2018-10-23 at 09 54 54

PR #933, opened 2018-10-23
Responsive Layout - Article teasers

Mitso:


Mitso:
@moh-moola 
Please review article teasers layout rearrangement

PR #936, opened 2018-10-24
Article template links bug

Mitso:


Mitso:
@moh-moola 
Please review article template update. 
Wagtail does not return Paragraph HTML entities if para value is specified on template.

PR #942, opened 2018-10-30
Responsive layout cleanup

Mitso:


Mitso:
@moh-moola Please review template and styles cleanup.

PR #946, opened 2018-10-31
Responsive Layout - Buttons and Headings

Mitso:


Mitso:
@moh-moola Please review. 
Template clean and buttons pages.

PR #947, opened 2018-11-01
Responsive Layout - Comments

Mitso:


Mitso:
@moh-moola please review.

PR #952, opened 2018-11-05
Responsive Layout - Comments pages

Mitso:


Mitso:
@moh-moola please review 

screenshot 2018-11-06 at 12 11 30

PR #957, opened 2018-11-08
Responsive Layout - Profiles Module

Mitso:


Mitso:
@moh-moola please review the FED related things.

PR #959, opened 2018-11-12
Responsive Layout - Surveys Component

Mitso:


Mitso:
@sewagodimo Please review.

PR #961, opened 2018-11-13
Update utils.py

codiebeulaine:


codiebeulaine:
the protocol for `request.site.root_url` is not HTTPS. since we have switched our QA sites to HTTPS, this needs to rely on the `wagtail_redirect_url`

PR #966, opened 2018-11-15
Responsive Layout - Yourwords Component

Mitso:


Mitso:
@moh-moola please review

PR #973, opened 2018-11-28
Add a managemenent command to remove empty nav_tags

sewagodimo:


gem/management/commands/remove_empty_nav_tags.py
```diff
@@ -0,0 +1,19 @@
...
+    def handle(self, **options):
+        for article in ArticlePage.objects.all():
+            for nav_tag in article.nav_tags.all():
+                if nav_tag.tag is None:
+                    nav_tag.delete()
+
+            if article.live:
+                article.save_revision().publish()
+            else:
+                article.save_revision()
```
KaitCrawford: I'm not suuuper happy that this is saving every article, regardless of if any tags were actually changed or not.

PR #974, opened 2018-11-29
remove_tags command: check for errors and only save changed files

sewagodimo:


gem/management/commands/remove_empty_nav_tags.py
```diff
@@ -11,9 +12,13 @@ def handle(self, **options):
...
-                article.save_revision().publish()
-            else:
-                article.save_revision()
+                    try:
+                        nav_tag.delete()
+                        if article.live:
+                            article.save_revision().publish()
+                        else:
+                            article.save_revision()
+                        print("nothing raised")
```
KaitCrawford: rogue print statement

gem/management/commands/remove_empty_nav_tags.py
```diff
@@ -11,9 +12,13 @@ def handle(self, **options):
...
             for nav_tag in article.nav_tags.all():
                 if nav_tag.tag is None:
-                    nav_tag.delete()
-
-            if article.live:
-                article.save_revision().publish()
-            else:
-                article.save_revision()
+                    try:
+                        nav_tag.delete()
```
KaitCrawford: I don't think this should be in the try block. Only the things that you expect to throw the error should be covered by it

gem/management/commands/remove_empty_nav_tags.py
```diff
@@ -11,9 +12,13 @@ def handle(self, **options):
...
+                        if article.live:
+                            article.save_revision().publish()
+                        else:
+                            article.save_revision()
+                        print("nothing raised")
+                    except IntegrityError:
+                            return ("IntegrityError: Only articles with sites "
+                                    "can be saved")
```
KaitCrawford: Please add something like the article id to this message. that way if we want to go and investigate further we know which articles are affected by this

gem/tests/test_management_commands.py
```diff
@@ -452,3 +453,16 @@ def test_command_works(self):
         # test that the article only points to existing tags
         self.assertEqual(article.nav_tags.count(), 1)
         self.assertEqual(article.nav_tags.all()[0].tag.title, 'New tag 2')
+
+        site = main.sites_rooted_here.all().first()
+        # deelte the second tag and the main's site
```
KaitCrawford: Spelling error

gem/tests/test_management_commands.py
```diff
@@ -452,3 +453,16 @@ def test_command_works(self):
...
+        tag2.delete()
+        site.delete()
+
+        cmd = main.sites_rooted_here.all().first()
+        cmd = call_command('remove_empty_nav_tags',)
```
KaitCrawford: Why do you assign this variable twice?

PR #985, opened 2019-01-11
upgrade to molo.surveys 8.0.3

sewagodimo:


requirements.txt
```diff
@@ -1,5 +1,6 @@
+molo.core==8.0.2
```
KaitCrawford: This is a duplicate requirement, molo.core is already required lower down.
Also this downgrades molo.core.

PR #986, opened 2019-01-15
upgrade molo-version

sewagodimo:


requirements.txt
```diff
@@ -1,5 +1,5 @@
 molo.surveys~=8.0.3
-molo.core==8.0.3
+molo.core~=8.0.4
```
KaitCrawford: I'm not sure I agree with making this dependency variable.
My feeling is that with something as integral to the project as molo.core is we should probably be explicit about which version we're using.
That's a subjective opinion though so I won't hold the PR to it

sewagodimo: @KaitCrawford Jamie actually said that this will make the docker image build fail. So I removed it, your intuition was correct

PR #995, opened 2019-01-24
Disable article moderation

KaitCrawford:


gem/templates/springster/wagtailadmin/pages/edit.html
```diff
@@ -0,0 +1,125 @@
+{% extends "wagtailadmin/base.html" %}
+{% load wagtailadmin_tags %}
+{% load gravatar %}
```
sewagodimo: @KaitCrawford it was here

PR #999, opened 2019-01-28
Add email TLS env variable

saeedmb8:


bearnard:
:+1: As long as the relevant libs support tls.

PR #1003, opened 2019-01-29
GEWEB 809: Article Teaser Navigation

Mitso:


Mitso:
@sewagodimo @codiebeulaine 
Please review.
Zathu, Base and Springster tags settings - template cleanup.

PR #1009, opened 2019-02-06
Final(Final) Quality Assurance

Mitso:


Mitso:
Update template image sizes for thumbnail and full images.

PR #1011, opened 2019-02-06
Translations Update

Mitso:


locale/rw/LC_MESSAGES/django.po
```diff
@@ -2419,10 +2278,7 @@ msgstr "Yasomwe:"
 
 #: gem/templates/rwanda/surveys/molo_survey_page.html:55
 #, fuzzy, python-format
```
codiebeulaine: should we remove this fuzzy?

PR #1021, opened 2019-02-13
WAK 95: Yegna Templates Header Module

Mitso:


Mitso:
@sewagodimo @moh-moola @saeedmb8  Initial proj setup.

Header module
screenshot 2019-02-14 at 13 06 45

PR #1060, opened 2019-03-05
Feature/add kaios manifest

codiebeulaine:


gem/views.py
```diff
@@ -280,6 +281,32 @@ def get(self, request, comment_pk):
...
 
+class KaiOSManifestView(View):
+    def get(self, request):
+        manifest = {
+            "version": "1.0.0",
+            "name": "Springster App",
+            "description": "An app providing information to girls",
+            "launch_path": "/",
+            "icons": {
+                "56": "/gem/statc/img/icons/next.png",
```
miltontony: This well probably be something like `/static/img/icons/next.png`

gem/views.py
```diff
@@ -280,6 +281,32 @@ def get(self, request, comment_pk):
         })
 
 
+class KaiOSManifestView(View):
+    def get(self, request):
```
miltontony: Almost.. you'll need to also change the headers as stipulated in the docs, so it might not necessarily be this method to use.

codiebeulaine: hmmm, ok

gem/views.py
codiebeulaine: ok

gem/views.py
```diff
@@ -280,6 +286,35 @@ def get(self, request, comment_pk):
...
+            },
+            "locales": {
+                "en": {
+                    "name": "Springster",
+                    "description": "An app providing information to girls"
+                }
+            },
+            "default_locale": "en"
+        }
+        response = HttpResponse(json.dumps(manifest))
```
miltontony:
```suggestion
        response = HttpResponse(json.dumps(manifest), content_type='application/x-web-app-manifest+json')
```

gem/views.py
```diff
@@ -280,6 +286,35 @@ def get(self, request, comment_pk):
...
+        response.headers['Content-Type'] = 'application/x-web-app-manifest+json'
```
miltontony:
```suggestion
```

PR #1061, opened 2019-03-05
Tablet Consolidated QA

Mitso:


Mitso:
@codiebeulaine @sewagodimo please review

PR #1062, opened 2019-03-06
WAK 69: Static template - update

Mitso:


Mitso:
@codiebeulaine @sewagodimo A review please

PR #1064, opened 2019-03-07
Consolidated Templates Cleanup - QA

Mitso:


Mitso:
@codiebeulaine Please review

Mitso:
Cleaned .po file for Yegna Templates and style padding.
Update the maintenance template file and error pages css.

PR #1067, opened 2019-03-08
QA Site - Update

Mitso:


gem/templates/yegna/core/article_page.html
```diff
@@ -53,7 +53,9 @@ <h1 class="heading__subheading heading__subheading--watch">{{ page.title}}</h1>
       {% if page.homepage_media %}
         {% for block in page.homepage_media %}
           {% with block.value as media %}
-            <p class="media-player__duration">({{media.duration|time:"i:s"}})</p>
+            {% if media.duration %}
+              <p class="media-player__duration">({{media.duration}})</p>
```
miltontony: I see you removed the duration format here. Was that intentional? i.e `time:"i:s"`

Mitso: @miltontony Yes it is. We need to find a working filter, this currently does not work. Codie is helping with this when she is available.

gem/templates/yegna/core/section_page-partials/section_video-article-teaser.html
```diff
@@ -45,7 +45,9 @@ <h1 class="teaser-{{teaser_type}}__title{% if article.title_highlight %} teaser-
     {% if article.homepage_media %}
       {% for block in article.homepage_media %}
         {% with block.value as media %}
-          <p class="teaser-{{teaser_type}}__duration">({{media.duration|time:"i:s"}})</p>
+          {% if media.duration %}
+            <p class="teaser-{{teaser_type}}__duration">({{media.duration}})</p>
```
miltontony: same as above re duration format

gem/templates/yegna/core/tags/hero_article.html
```diff
@@ -73,7 +73,9 @@ <h1 class="teaser-stacklist__title{% if article.title_highlight %} teaser-stackl
               {% if article.homepage_media %}
                 {% for block in article.homepage_media %}
                   {% with block.value as media %}
-                    <p class="teaser-stacklist__duration">({{media.duration}})</p>
+                    {% if media.duration %}
+                      <p class="teaser-stacklist__duration">({{media.duration}})</p>
```
miltontony: same question regarding duration format

gem/templates/yegna/core/article_page.html
miltontony: 👌

PR #1072, opened 2019-03-13
WAK 246: FED Ethiopia Post Launch

Mitso:


gem/styles/gem-yegna/modules/_m-article-teaser-stacklist.scss
```diff
@@ -85,13 +85,15 @@
...
         max-height: 422px;
       }
+      @media screen and (min-width: 72.188em) {
+        width: 785px;
+      }
     }
     #{$block}__thumbnail {
       @media screen and (min-width: 62.000em)  {
-        min-width: 750px;
+        min-width: 785px;
```
sewagodimo: is this min value used in other templates? 
Might be worth making a variable

Mitso: Thanks Sewa. It's not used elsewhere or even in a consistent manner atm so no.

PR #1091, opened 2019-03-28
WAK 252: D-Pad Nav Implementation

Mitso:


gem/static/js/kaios.js
```diff
@@ -0,0 +1,70 @@
...
+        case 'ArrowLeft':
+          nav(-1);
+          break;
+      }
+    };
+    document.activeElement.addEventListener('keydown', handleKeydown);
+  };
+
+  var softkeyCallback = {
+    left: function() { /*console.log('You click on SoftLeft')*/ },
```
codiebeulaine: can you uncomment these console.logs?

PR #1093, opened 2019-04-02
WAK 260: Optimise GEM ui Apps

Mitso:


gem/static/js/kaios.js
```diff
@@ -5,43 +5,48 @@
...
           break;
         case 'ArrowRight':
           nav(1);
-          console.log('You click on ArrowRight')
+          console.log('You click on ArrowRight', e.which || e.keyCode || 0)
           break;
         case 'ArrowLeft':
           nav(-1);
-          console.log('You click on ArrowLeft')
+          console.log('You click on ArrowLeft', e.which || e.keyCode || 0)
```
sewagodimo: Should it not be "you click**ed** on.."

PR #1103, opened 2019-04-10
add setting to enable chat bot in admin

moh-moola:


gem/models.py
```diff
@@ -125,6 +128,8 @@ class GemSettings(BaseSetting):
             ],
             heading="BBM",
         ),
+        MultiFieldPanel(
+            [FieldPanel('fb_enable_chat_bot')], heading="Face Book")
```
codiebeulaine: Can we make Facebook one word?

codiebeulaine:
left a tiny comment but 👍 when travis passes

gem/models.py
moh-moola: Just did

PR #1108, opened 2019-04-15
Templates tags update

Mitso:


Mitso:
@moh-moola @codiebeulaine Please review

Cleanup Templates to always check for Non-English Main Language word titles in English.

PR #1114, opened 2019-04-23
change clashing username when updating user fields

codiebeulaine:


gem/backends.py
```diff
@@ -39,23 +39,36 @@ def _update_user_from_claims(user, claims):
...
 
     # Ensure the profile is linked to their auth service account using the uuid
     if user.profile.auth_service_uuid is None:
         user.profile.auth_service_uuid = claims.get("sub")
 
+        # If a user already exists with this username
+        # change that user's username
+        username = claims.get("username")
+        if username:
+            for user in User.objects.filter(username=username):
```
KaitCrawford: The variable `user` is already being used before here. I suggest picking a different name for this variable.

gem/backends.py
```diff
@@ -39,23 +39,36 @@ def _update_user_from_claims(user, claims):
...
+                user.username = user.profile.site.pk + '_' + username
```
KaitCrawford: I think we only want to change the user name if the users that are returned are not the current user.
Also, was the decision to just add the site id to the username? Is that guaranteed to be unique or is there some case where it might clash as well?

gem/backends.py
```diff
@@ -39,23 +39,36 @@ def _update_user_from_claims(user, claims):
...
     if date_of_birth:
         date_of_birth = datetime.strptime(date_of_birth, "%Y-%m-%d").date()
     user.profile.date_of_birth = date_of_birth
     if user.profile.alias is None or user.profile.alias == "":
-        user.profile.alias = user.username
-    elif user.profile.alias != user.username:
-        user.profile.alias = user.username
+        user.profile.alias = user.profile.migrated_username
+    elif user.profile.alias != user.profile.migrated_username:
+        user.profile.alias = user.profile.migrated_username
```
KaitCrawford: I think we should still set these to their new username. We don't change their migrated username before this do we?
So what this is basically doing is setting their alias to their old username. but what if they choose a new username that's vastly different to their old one? Then their alias is going to be wrong

gem/backends.py
```diff
@@ -39,23 +39,36 @@ def _update_user_from_claims(user, claims):
...
     user.profile.save()
+    if username:
+        user.username = username
+        user.save()
```
KaitCrawford: I think this block should be up at the top where we change the clashing username.
That way it should only happen when users are reclaiming accounts and not on every login.

gem/backends.py
```diff
@@ -148,6 +161,13 @@ def create_user(self, claims):
         username = claims.get("preferred_username")
         email = claims.get("email", "")  # Email is optional
         # We create the user based on the username and optional email fields.
+
+        # If a user already exists with this username
+        # change that user's username
+        for user in self.UserModel.objects.filter(username=username):
+            user.username = user.profile.site.pk + '_' + username
+            user.save()
```
KaitCrawford: I think we should only do this if the user that we're clashing with doesn't have an `auth_service_uuid`.
And if they do then we need to raise an error because there's something very wrong.

gem/backends.py
```diff
@@ -36,15 +37,29 @@ def _update_user_from_claims(user, claims):
...
     # Ensure the profile is linked to their auth service account using the uuid
     if user.profile.auth_service_uuid is None:
         user.profile.auth_service_uuid = claims.get("sub")
 
+        # If a user already exists with this username
+        # change that user's username
+        if username:
+            for u in User.objects.filter(
+                    username=username).exclude(pk=user.pk):
+                u.username = str(u.profile.site.pk) + '_' + username
```
KaitCrawford: We should probably also check that this user doesn't already have an `auth_service_uuid` and throw the `FieldError` here if they do

gem/backends.py
```diff
@@ -148,6 +163,16 @@ def create_user(self, claims):
         username = claims.get("preferred_username")
         email = claims.get("email", "")  # Email is optional
         # We create the user based on the username and optional email fields.
+
+        # If a user already exists with this username
+        # change that user's username
+        for user in self.UserModel.objects.filter(username=username):
+            if user.profile.auth_service_uuid is None:
```
KaitCrawford: Something we encountered previously is that some users don't have profiles (I'm guessing this is admin or QA users)
Any users like that will cause an error here. you might need to catch it or check that they have a profile beforehand.

gem/backends.py
```diff
@@ -148,6 +163,16 @@ def create_user(self, claims):
...
+                user.username = str(user.profile.site.pk) + '_' + username
+                user.save()
+            else:
+                raise FieldError
```
KaitCrawford: Is there any way to add a message to this error? If there is then I think we should add some sort of explanation of why we're raising this error.

gem/backends.py
```diff
@@ -36,15 +37,34 @@ def _update_user_from_claims(user, claims):
...
+        if username:
+            for u in User.objects.filter(
+                    username=username).exclude(pk=user.pk):
+                if user.profile.auth_service_uuid is None:
+                    user.username = str(user.profile.site.pk) + '_' + username
+                    user.save()
+                else:
+                    raise FieldError(
+                        'Desired username clashes with user with pk %s whose'
+                        ' profile has no auth_service_uuid' % u.pk)
```
KaitCrawford: Typo? If we get here then the user that they are clashing with does have an `auth_service_uuid`

gem/backends.py
```diff
@@ -36,15 +37,34 @@ def _update_user_from_claims(user, claims):
...
         user.profile.auth_service_uuid = claims.get("sub")
 
+        # If a user already exists with this username
+        # change that user's username
+        if username:
+            for u in User.objects.filter(
+                    username=username).exclude(pk=user.pk):
+                if user.profile.auth_service_uuid is None:
+                    user.username = str(user.profile.site.pk) + '_' + username
+                    user.save()
```
KaitCrawford: You're using the wrong variable throughout here. `u` is the user that we're clashing with, we want to be changing their details. Right now you're checking and changing the details for the person logging in.

gem/backends.py
```diff
@@ -148,6 +168,18 @@ def create_user(self, claims):
...
+        # If a user already exists with this username
+        # change that user's username
+        for user in self.UserModel.objects.filter(username=username):
+            if user.profile.auth_service_uuid is None:
+                user.username = str(user.profile.site.pk) + '_' + username
+                user.save()
+            else:
+                raise FieldError(
+                        'Desired username clashes with user with pk %s whose'
+                        ' profile has no auth_service_uuid' % user.pk)
```
KaitCrawford: Same typo here I think?

gem/tests/test_auth.py
```diff
@@ -221,6 +249,30 @@ def test_update_user_from_claims_creates_profile(self):
...
+        user2 = get_user_model().objects.get(id=user2.pk)
+        self.assertFalse(user.is_superuser)
+        self.assertEqual(user.first_name, 'testgivenname')
+        self.assertEqual(user.last_name, 'testfamilyname')
+        self.assertEqual(user.email, 'test@email.com')
+        self.assertEqual(user2.username, '3_john')
+        self.assertEqual(user.username, 'john')
+        self.assertEqual(
+            str(user.profile.auth_service_uuid),
+            'e2556752-16d0-445a-8850-f190e860dea4')
```
KaitCrawford: You should probably add a test for the FieldError that you are raising in `_update_user_from_claims`

PR #1116, opened 2019-04-25
WAK 343: Springster Templates JS

Mitso:


Mitso:
@moh-moola @saeedmb8 Please review 

An auth template theming update.
JS inside if is_via_freebasics check

PR #1130, opened 2019-05-16
add videospage model to gem models

codiebeulaine:


gem/models.py
```diff
@@ -15,6 +17,28 @@
 from wagtail.images.edit_handlers import ImageChooserPanel
 
 
+class VideosPage(Page):
+    date = models.DateField("Post date")
+    intro = models.CharField(max_length=250)
+    body = RichTextField(blank=True)
+    videourl = models.CharField(max_length=250, null=True)
```
codiebeulaine: Django has a URLField https://docs.djangoproject.com/en/1.11/ref/models/fields/#urlfield

(Just a note, we use Django 1.11.x)

gem/models.py
```diff
@@ -15,6 +17,28 @@
 from wagtail.images.edit_handlers import ImageChooserPanel
 
 
+class VideosPage(Page):
+    date = models.DateField("Post date")
```
codiebeulaine: I have also made a previous comment asking why this date field is necessary?

gem/models.py
```diff
@@ -15,6 +17,28 @@
 from wagtail.images.edit_handlers import ImageChooserPanel
 
 
+class VideosPage(Page):
```
codiebeulaine: If this were to inherit from ArticlePage (molo.core.ArticlePage), then all these field would already be here. This model doesn't offer any functionality that the ArticlePage doesn't already offer.

PR #1131, opened 2019-05-17
GA Parameters: article tags (hashtags)

sewagodimo:


gem/middleware.py
```diff
@@ -54,6 +55,25 @@ class GemMoloGoogleAnalyticsMiddleware(MoloGoogleAnalyticsMiddleware):
...
+        path_components = [component for component in path.split('/')
+                           if component]
+
+        try:
+            page, args, kwargs = request.site.root_page.specific.route(
+                request,
+                path_components)
+            if issubclass(type(page), ArticlePage):
+                tags = page.tags_list()
+                print('|'.join(tags))
```
miltontony: stray print

gem/middleware.py
```diff
@@ -54,6 +55,25 @@ class GemMoloGoogleAnalyticsMiddleware(MoloGoogleAnalyticsMiddleware):
...
+        path = request.get_full_path()
+        path_components = [component for component in path.split('/')
+                           if component]
+
+        try:
+            page, args, kwargs = request.site.root_page.specific.route(
+                request,
+                path_components)
+            if issubclass(type(page), ArticlePage):
+                tags = page.tags_list()
```
miltontony: Do you not need `page.specific.tags_list()`? Just wondering if this is working..

miltontony: Also, is there a chance that `page.tags_list()` will return `None`? Because that will be a problem for the join

gem/middleware.py
```diff
@@ -54,6 +55,25 @@ class GemMoloGoogleAnalyticsMiddleware(MoloGoogleAnalyticsMiddleware):
...
+                print('|'.join(tags))
+            return '|'.join(tags)
+
+        except Http404:
+            raise []
```
miltontony: What are you raising here? did you intend maybe to `return` instead of `raise`?

sewagodimo: @miltontony yeap, it's suppose to be a return

gem/middleware.py
sewagodimo: it will return  an empty arrayset [] when there a are no tags
`if issubclass(type(page), ArticlePage):` will ensure that it's always an ariclepage and the tags_list() will always return an empty list when there are no tags

gem/middleware.py
sewagodimo:
```suggestion
            return []
```

gem/middleware.py
sewagodimo:
```suggestion
```

gem/middleware.py
miltontony: I think the question is whether `route()` returns a concrete type (i.e Page vs ArticlePage). The only way to be sure is to call `page.specific`

gem/middleware.py
```diff
@@ -54,6 +55,24 @@ class GemMoloGoogleAnalyticsMiddleware(MoloGoogleAnalyticsMiddleware):
...
+        tags = []
+        path = request.get_full_path()
+        path_components = [component for component in path.split('/')
+                           if component]
+
+        try:
+            page, args, kwargs = request.site.root_page.specific.route(
+                request,
+                path_components)
+            if issubclass(type(page), ArticlePage):
```
miltontony:
```suggestion
            if issubclass(type(page.speficic), ArticlePage):
```

gem/middleware.py
miltontony:
```suggestion
                tags = page.specific.tags_list()
```

gem/middleware.py
```diff
@@ -54,6 +55,24 @@ class GemMoloGoogleAnalyticsMiddleware(MoloGoogleAnalyticsMiddleware):
...
+        path = request.get_full_path()
+        path_components = [component for component in path.split('/')
+                           if component]
+
+        try:
+            page, args, kwargs = request.site.root_page.specific.route(
+                request,
+                path_components)
+            if issubclass(type(page.specific), ArticlePage):
+                tags = page.tags_list()
```
sewagodimo:
```suggestion
                tags = page.specific.tags_list()
```

PR #1132, opened 2019-05-17
Auth Theme Styles - update

Mitso:


Mitso:
@codiebeulaine @KaitCrawford 

Please review generated Core Auth theme styles.
This is generated here as per the previous set-up and having no access to auth repos.

PR #1133, opened 2019-05-20
Chhaajja base UI

codiebeulaine:


gem/settings/base.py
```diff
@@ -37,31 +37,31 @@
 # Authentication Service Tokens
 USE_OIDC_AUTHENTICATION = environ.get(
     'USE_OIDC_AUTHENTICATION', '') == 'true'
-OIDC_RP_CLIENT_ID = "unused"  # Some constructors require that this be set.
-OIDC_RP_CLIENT_SECRET = "unused"  # some constructors require that this be set.
+OIDC_RP_CLIENT_ID = "118274"  # Some constructors require that this be set.
```
codiebeulaine: the client ID and SECRET does not need to be here, it is set in the django-admin. In future, whern these are uploaded we need to change them and create new ones since github is public

gem/settings/base.py
```diff
@@ -216,7 +216,10 @@ def get_default_template(site_layout_base, site_layout):
     }
 
 
-SITE_LAYOUT_BASE = environ.get('SITE_LAYOUT_BASE', 'base')
+SITE_LAYOUT_BASE = environ.get('SITE_LAYOUT_BASE', 'chhaajja')
```
codiebeulaine: can we please rename the templates to match the chhaa-jaa, there are too many different spellings of Chhaa Jaa

gem/settings/dev.py
```diff
@@ -4,7 +4,8 @@
 ALLOWED_HOSTS = [
     'localhost',
     '.localhost',
-    '127.0.0.1'
+    '127.0.0.1',
+    'chajaa.redmatter.tech'
```
codiebeulaine: this doesn't need to be added here as it is only for you local settings

gem/settings/base.py
```diff
@@ -37,31 +37,31 @@
...
+OIDC_OP = environ.get('OIDC_OP', 'https://authentication-service.qa-hub.ie.gehosting.org')
+THEME = environ.get('THEME', 'chhaa-jaa')
+LOGIN_REDIRECT_URL = environ.get('LOGIN_REDIRECT_URL', 'https://chajaa.redmatter.tech/')
 LOGIN_URL = 'molo.profiles:auth_login'
 LOGOUT_URL = 'molo.profiles:auth_logout'
 REGISTRATION_URL = reverse_lazy('molo.profiles:user_register')
 VIEW_PROFILE_URL = reverse_lazy('molo.profiles:view_my_profile')
 EDIT_PROFILE_URL = reverse_lazy('edit_my_profile')
-LOGOUT_REDIRECT_URL = environ.get('LOGOUT_REDIRECT_URL', '')
+LOGOUT_REDIRECT_URL = environ.get('LOGOUT_REDIRECT_URL', 'https://chajaa.redmatter.tech/')
```
codiebeulaine: all of these settings are for you to use locally, I have updated the spinnaker config to reflect these as environment variables, but chajaa.redmatter.tech is your local instance and not the actual QA site

PR #1142, opened 2019-05-27
Add gem navigation tags to GA custom params [cd6] on articlepage view

sewagodimo:


codiebeulaine:
@sewagodimo we already have code that gets the tags in molo core, maybe would be better to call that code? It would be in template tags

codiebeulaine:
if not then this code should probably exist in molo core middleware?

PR #1151, opened 2019-05-29
WAK 410: Article Image backup

Mitso:


Mitso:
Screenshot 2019-05-30 at 00 21 14

gem/templates/chhaajaa/core/tags/hero_article.html
```diff
@@ -7,31 +7,31 @@
...
-              {% for block in article.body %}
+            {% for block in article.body %}
+              {% if block.value.type == "video" %}
                 {% if block.value.youtube_link %}
                   <div class="teaser-stacklist__iframe">
                     {% embed block.value.youtube_link %}
                   </div>
+                  {% endif %}
+              {% else %}
+                {% if forloop.last and article.get_effective_image %}
```
codiebeulaine: won't this display the video and the image? Like if it gets a block that is a video it will display it, and then if the next block isn't a video but is the last in the loop, it will display the image?

gem/templates/chhaajaa/core/tags/hero_article.html
```diff
@@ -99,5 +99,93 @@ <h4 class="teaser-stacklist__subtitle">{{ article.subtitle|smarttruncatechars:15
...
+          e.target.closest(".comments__form-post").setAttribute("style","border: 5px solid #72f3a2; outline: none; padding: 10px; border-radius: 5px;");
+        } else {
+          console.log(document.activeElement);
+          //e.target.closest(".comments__form-post").setAttribute("style");
+        }
+      });
+
+      var elems  = document.getElementsByClassName("choice-group");
+      for(var i = 0; i < elems.length; i++) {
+        console.log(elems[i]);
```
KaitCrawford: Did you forget to remove these print statements? or are they needed for something?

PR #1152, opened 2019-05-30
Add submit comment via ajax for chhaa jaa

codiebeulaine:
This PR overrides the submit call for a comment within the Chhaa Jaa template and submits the comment via ajax. If the comment was submitted successfully it removes the comment text from the comment text area. Since Chhaa Jaa only recevies comments and does not display it anywhere, no further work is needed.

### DIFF
--- a/gem/templates/chhaajaa/comments/comment_block.html
+++ b/gem/templates/chhaajaa/comments/comment_block.html
@@ -8,14 +8,6 @@
       {% if article.allow_commenting %}
         {% if request.user.is_authenticated %}
           {% render_comment_form for article %}
-          {% comment %}<!--
-          {% else %}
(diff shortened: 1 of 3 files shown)


gem/templates/chhaajaa/comments/comment_block.html
```diff
@@ -8,14 +8,6 @@ <h4 class="heading__subheading">{% trans "Comment" %}</h4>
       {% if article.allow_commenting %}
         {% if request.user.is_authenticated %}
           {% render_comment_form for article %}
-          {% comment %}<!--
-          {% else %}
```
codiebeulaine: this else does need need to be here because in chhaa jaa users are not able to see any content before they have logged in, so the user is always authenticated

PR #1157, opened 2019-05-30
initial articel page

codiebeulaine:


codiebeulaine:
Screenshot 2019-05-30 at 13 45 22

gem/templates/chhaajaa/core/article_page.html
```diff
@@ -1,342 +1,187 @@
...
+              //this.classList.add('selected');
+            } else {
+              this.classList.add('selected');
+              //el.classList.remove('selected');
+            }
+          });
+        }
+  
+        //COPIED SCRIPT
+        const Utils = {
```
Mitso: You can remove this block, only for the homepage excerpt.

PR #1171, opened 2019-05-31
QA Review & Cleanup

Mitso:


gem/settings/dev.py
```diff
@@ -5,6 +5,7 @@
     'localhost',
     '.localhost',
     '127.0.0.1',
+    '192.168.101.242'
```
codiebeulaine: do we wana land with this?

PR #1173, opened 2019-06-01
Templates cleanup

Mitso:


gem/templates/chhaajaa/core/tags/section_listing_homepage.html
```diff
@@ -6,14 +6,14 @@
             work 3
         {% content_is section "News" as is_watch_news %}
         {% if is_watch_news %}
-          
+
           {% load_descendant_articles_for_section section featured_in_homepage=True as articles %}
-            work4 
+            work4
```
codiebeulaine: work4? I don't think we need news here?

Mitso: I'm actually not sure who and when this was added, but didn't delete it.

PR #1175, opened 2019-06-03
WAK 433: Responsive profiles and buttons / QA - review

Mitso:


Mitso:
Video Iframe padding. 
Comments icon cta and clicking to Comments title.
Comments enable / disabled setting FED ui

PR #1178, opened 2019-06-03
add correct videos to side bar

codiebeulaine:


gem/templates/chhaajaa/core/section_page.html
```diff
@@ -180,7 +180,13 @@ <h6 class="teaser-card__subtitle">
         {% endif %}
       </div>
 
+      {% comment %} <!-- Beginning of Side Bar --> {% endcomment %}
+      {% content_is self "Videos" as is_video_section %}
+      {% if not is_video_section %}
+        {% load_child_articles_for_section self.get_first_sibling.specific as articles %}
```
Mitso: @codiebeulaine Whats the use of the  `load_child_articles_for_section` here if the section articles are not video articles?

codiebeulaine: @Mitso the section I am giving it is the sibling of the current section, so I'm calling this template tag but with a different section

Mitso: Ok cool, thanks @codiebeulaine 👍

PR #1219, opened 2019-06-21
add script for the content content languages

sewagodimo:


gem/management/commands/change_content_language.py
```diff
@@ -0,0 +1,37 @@
...
+                        "Invalid language primary keys were entered" +
+                        str(old_lang_pk) + " and " + str(new_lang_pk)))
+                else:
+                    pages = models.MoloPage.objects.all()
+                    for page in pages:
+                        if hasattr(page.specific, 'language'):
+                            if page.specific.language == old_lang:
+                                page.specific.language = new_lang
+                                page.specific.save()
+                                page.save()
```
codiebeulaine: you might need to do save_revision().publish() if the article or page is live and save_revision() if it is draft?

sewagodimo: thank you

sewagodimo: @codie I tested it and the language change doesn't require me to publish the pages again, it actually makes the changes fail for some of the content

codiebeulaine:
👍

PR #1227, opened 2019-06-28
add a hyperlink to page titles

sewagodimo:


gem/styles/gem-rwanda/modules/_m-article-teasers-tags.scss
```diff
@@ -25,6 +25,11 @@
     font-size: $font-small;
     text-decoration: underline;
     text-underline-position: under;
+    .base-teaser__title{
```
Mitso: @sewagodimo Please locate which UI that has the classes - article teasers tags. 
And then check where this element picking up these styles rules

gem/styles/gem-rwanda/modules/_m-article-teasers.scss
```diff
@@ -108,4 +108,10 @@
       }
     }
   }
+  .base-teaser__title{
```
Mitso: @sewagodimo Please locate which UI that has the classes. 
And then check where this element picking up these styles rules.

PR #1240, opened 2019-07-08
add script for moving survey models excluding personalisable models

codiebeulaine:


gem/management/commands/migrate_surveys_to_forms.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class Command(BaseCommand):
+    def handle(self, *args, **options):
+        for main in Main.objects.all():
+            surveys_index = SurveyIndexPage.objects.child_of(main).first()
+            forms_index = FormIndexPage.objects.child_of(main).first()
+            for survey in MoloSurveyPage.objecs.descendent_of(surveys_index):
+                survey_dict = survey.__dict__
+                del survey_dict['_state']
```
KaitCrawford: What is stored in the `_state` attribute?

gem/management/commands/migrate_surveys_to_forms.py
```diff
@@ -0,0 +1,36 @@
...
+                form = MoloFormPage.objects.create(**survey_dict)
+                forms_index.add_child(form)
+
+                for form_field in MoloSurveyFormField.objects.filter(
+                        page__pk=survey.pk):
+                    form_field_dict = form_field.__dict__
+                    del form_field_dict['_state']
+                    MoloFormsFormField.objects.create(**form_field_dict)
```
KaitCrawford: this is another place where we're going to have issues if the pk changes. we'll have to make sure that the fields point to the new survey, not the old one.

gem/management/commands/migrate_surveys_to_forms.py
```diff
@@ -0,0 +1,36 @@
...
+    MoloFormPage, FormIndexPage, MoloFormView,
+    MoloFormsFormField, MoloFormPageSubmission)
+from molo.core.models import Main
+
+
+class Command(BaseCommand):
+    def handle(self, *args, **options):
+        for main in Main.objects.all():
+            surveys_index = SurveyIndexPage.objects.child_of(main).first()
+            forms_index = FormIndexPage.objects.child_of(main).first()
```
KaitCrawford: Don't we have these indexes for all the sites? So we'll want to change it for more than just 'Main' right?
and are there different index pages for different languages or do they use the same one?

PR #1242, opened 2019-07-08
Update the comment's view's page title to flag that it's a tile

sewagodimo:
When viewing the comments of an article page, thecomments page tile is the same as that of the article page. For example: http://za.heyspringster.com/sections/my-life/do-you-feel-alone-in-the-world/ has the same page title as it’s comments view: http://za.heyspringster.com/commenting/molo/48/comments/

This is causing a problem for web crawlers.

Proposed solution:
Add ‘Comments -’ at the beginning of the page title when the comments are being displayed

### DIFF
--- a/gem/templates/springster/comments/comments.html
+++ b/gem/templates/springster/comments/comments.html
@@ -1,5 +1,6 @@
 {% extends "core/article_page.html" %}
 {% load wagtailcore_tags mptt_tags molo_commenting_tags gem_tags %}
+{% block title %}Comments - {{ self.title}}{% endblock %}
 {% block content %}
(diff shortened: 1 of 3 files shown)


gem/templates/springster/comments/comments.html
```diff
@@ -1,5 +1,6 @@
 {% extends "core/article_page.html" %}
 {% load wagtailcore_tags mptt_tags molo_commenting_tags gem_tags %}
+{% block title %}Comments - {{ self.title}}{% endblock %}
```
Mitso: @sewagodimo This might need to be translated.

PR #1260, opened 2019-07-15
WAK 591: Nav review

Mitso:


Mitso:
Update Section nav active states across articles or other pages.

PR #1281, opened 2019-07-29
WAK 662: Springster homepage  

Mitso:


gem/templates/springster/core/main.html
```diff
@@ -30,38 +30,24 @@
...
-    {% include 'core/tags/latest_listing_homepage.html' with articles=latest_articles|slice:"9:12" %}
-
-    {% include "patterns/components/article-teasers/sp_variations/carousel-teaser-listing.html" with tag=tag_nav_data.tags_list.1.0 tag_articles=tag_nav_data.tags_list.1.1 %}
+    {% bannerpages position=1 %}
 
-    {% bannerpages position=3 %}
-    {% your_words_competition %}
+    {% comment %}
+      <!-- TAG CLOUD | ON OTHER PAGES -->
+    {% endcomment %}
```
saeedmb8: @Mitso do we need this comment here?

Mitso: @saeedmb8 
Yes. The idea is that to start commenting decisions made on templates components going forward.

This Django & HTML commenting tags means this is only viewed on the template level so it's useful for us.

PR #1285, opened 2019-08-06
Open Graph fb share params update

Mitso:


gem/templates/springster/base.html
```diff
@@ -61,6 +65,40 @@
...
+              for(let j = 0; j < formats[i].ext.length; j++) {
+                if(formats[i].ext[j] == imageExt) {
+                  document.querySelector("meta[property='og:type']").setAttribute("content",formats[0].type + "/" + formats[0].ext[i]);
+                }
+              }
+            }
+          }
+
+    </script>
   </head>
```
moh-moola: @Mitso I can create a django template filter for this, seeing that we are having more and more media content that might require it
```
import mimetypes    

>>> mimetypes.guess_type('a.py',strict = True)
('text/x-python', None)
```

PR #1352, opened 2020-01-29
GE 21 1012 - Tujibebe & Chhaajaa Meta Tags

Mitso:


gem/templates/tanzania/base.html
```diff
@@ -13,7 +13,7 @@
 <!--<![endif]-->
   <head>
     <meta charset="utf-8"/>
-    <title>{% block title %}{% if self.seo_title %}{{ self.seo_title }}{% else %}{{ self.title }}{% endif %}{% endblock %}</title>
+    <title>Tujibebe | {% if self.seo_title %} {% if self.seo_title == "Main" %} Home {% else %}{{ self.seo_title }}{% endif %}{% else %} {% if self.title == "Main" %} Home {% else %}{{ self.title }}{% endif %}{% endif %}</title>
```
moh-moola: @Mitso the seo page titles don't the get translated?

moh-moola: if logic blocks have been changed to if blocks...

Mitso: Thanks @moh-moola , the dynamic content get the current page translation from the CMS.

moh-moola: @Mitso I think this should be spread across multiple lines, it makes it easier to understand and see logical errors

PR #1358, opened 2020-02-11
add allauth and enable google social login for admins

moh-moola:


Mitso:
@moh-moola 

Please see update 
google-entry

gem/tests/test_auth.py
```diff
@@ -441,3 +447,291 @@ def test_auth_middleware(self):
...
+            title='main1', slug='main1',
+            path='00010002', url_path='/main1/'
+        )
+        self.site = self.main.get_site()
+
+    @override_settings(ENABLE_ALL_AUTH=True)
+    def test_admin_login_view(self):
+        res = self.client.get(reverse('wagtailadmin_login'))
+        self.assertEqual(res.status_code, 200)
+        self.assertEqual(settings.ENABLE_ALL_AUTH, True)
```
KaitCrawford: Since you set this setting before the test you probably don't need to check it here.
What is this test actually checking? that the login view doesn't redirect? Is the user already authenticated?
Maybe you should be checking which url we land at or which template is used?

gem/tests/test_auth.py
```diff
@@ -441,3 +447,291 @@ def test_auth_middleware(self):
...
+
+    def test_admin_login_view_default_allauth_setting(self):
+        res = self.client.get(reverse('wagtailadmin_login'))
+        self.assertEqual(res.status_code, 200)
+        self.assertEqual(settings.ENABLE_ALL_AUTH, False)
```
KaitCrawford: Same comment as above.

gem/tests/test_auth.py
```diff
@@ -441,3 +447,291 @@ def test_auth_middleware(self):
...
+        self.assertEqual(res.url, '/admin/')
+
+        res = self.client.get(res.url)
+        self.assertEqual(res.status_code, 200)
+
+    @override_settings(ENABLE_ALL_AUTH=False)
+    def test_login_all_auth_disabled(self):
+        res = self.client.get(reverse('wagtailadmin_login'))
+        self.assertEqual(res.status_code, 200)
+        self.assertEqual(settings.ENABLE_ALL_AUTH, False)
```
KaitCrawford: This test is pretty much exactly the same as `test_admin_login_view_default_allauth_setting`
Did you mean to log the user in first?

gem/tests/test_auth.py
```diff
@@ -441,3 +447,291 @@ def test_auth_middleware(self):
...
+        req.site = self.site
+        req.user = self.user
+
+        invite = Invite.objects.create(
+            email=data['email'], user=self.user, site=self.site)
+        self.client.force_login(self.user)
+
+        url = '/admin/gem/invite/edit/{}/'.format(invite.pk)
+        res = self.client.post(url, request=req)
+        self.assertEqual(res.status_code, 200)
```
KaitCrawford: Please test that the existing invite is displayed or that the correct template is being used here.

moh-moola:
@KaitCrawford I have made amendments to the tests

gem/tests/test_auth.py
```diff
@@ -441,3 +447,289 @@ def test_auth_middleware(self):
...
+        self.site = self.main.get_site()
+
+    @override_settings(ENABLE_ALL_AUTH=True)
+    def test_admin_login_view(self):
+        res = self.client.get(reverse('wagtailadmin_login'))
+        self.assertEqual(res.status_code, 200)
+
+    def test_admin_login_view_default_allauth_setting(self):
+        res = self.client.get(reverse('wagtailadmin_login'))
+        self.assertEqual(res.status_code, 200)
```
KaitCrawford: What is this test and the one above it actually supposed to be testing? You didn't make any changes to the `wagtailadmin_login` url
I presume this test and the one above it are supposed to load different pages but these tests don't make that clear. Please add a check to each of them to show they are loading the correct templates or content

gem/tests/test_auth.py
```diff
@@ -441,3 +447,299 @@ def test_auth_middleware(self):
...
+        )
+        self.site = self.main.get_site()
+
+    @override_settings(ENABLE_ALL_AUTH=True)
+    def test_admin_login_view(self):
+        res = self.client.get(reverse('wagtailadmin_login'))
+        self.assertEqual(res.status_code, 200)
+        # toDo: find a better way to handle conditional urls patterns,
+        #  because django only loads them once on server instantiation
+        # self.assertTemplateUsed(res, 'wagtailadmin/social_login.html')
```
moh-moola, line 468: @KaitCrawford we need to find a better way to handle conditional urls patterns, because django only loads them once on server instantiation

KaitCrawford: ok

PR #1366, opened 2020-03-27
use request language in template

moh-moola:
The idea here is to use the article language as the default for new session, so article are not translated to the main lang


gem/middleware.py
```diff
@@ -41,6 +43,24 @@ def process_request(self, request):
             del request.META['HTTP_ACCEPT_LANGUAGE']
 
 
+class GemLocaleMiddleware(LocaleMiddleware):
+    def process_request(self, request):
+        has_slug = len(request.path.split('/')) > 1
```
KaitCrawford, line 48: shouldn't this be `>3` because the request has quite a few components? eg https://springster.qa-hub.ie.gehosting.org/sections/my-body/how-healthy-are-you/

moh-moola: this is a safety mechanism, so long as the url is not the root, check for the slug for potential url config changes. there is also a check to make sure the url is present to do the lookup

PR #1368, opened 2020-04-06
Feature/ge21 1148 site perms

moh-moola:


gem/tests/test_middleware.py
```diff
@@ -448,3 +449,79 @@ def test_non_default_language_new_session(self):
...
+            model='admin'
+        )
+        admin_permission, created = Permission.objects.get_or_create(
+            content_type=wagtailadmin_content_type,
+            codename='access_admin',
+            name='Can access Wagtail admin'
+        )
+        return admin_permission
+
+    def setUp(self):
```
KaitCrawford, line 467: I think a bunch of this setup can be removed. You don't need multiple languages, the section or the articles, do you?

moh-moola: @KaitCrawford 👍

PR #1370, opened 2020-04-14
update molo version to molo version 10.2 which deprecates reaction qu…

moh-moola:
…estions


gem/tests/test_views.py
```diff
@@ -626,72 +624,3 @@ def test_extra_params_set(self):
         self.assertEqual(view.get_extra_params(request), {
             'theme': 'springster', 'language': 'en'
         })
-
-
-class ChhaaJaaReactionQuestionsTest(TestCase, GemTestCaseMixin):
```
KaitCrawford, line 631: Please replace this test with one that checks the same things for the new forms reaction questions

gem/templates/malawi/core/related-article-list.html
```diff
@@ -0,0 +1,69 @@
+{% load wagtailcore_tags wagtailimages_tags core_tags %}
+
```
KaitCrawford, line 2: Why is this template part of this PR? it doesn't seem to be related to any of this work.

moh-moola: @KaitCrawford not sure as to why this was not causing errors before, but the omitting this template will cause 500 errors

KaitCrawford: @moh-moola was it raising errors on your local instance or in the tests?

moh-moola: @KaitCrawford locally

gem/tests/test_views.py
moh-moola: @KaitCrawford done

gem/tests/test_views.py
```diff
@@ -637,15 +643,68 @@ def setUp(self):
...
+            )
+
+            res = self.client.get('/')
+            self.assertEqual(res.status_code, 200)
+            self.assertContains(res, article.title)
+            self.assertTrue(form.title in str(res.content))
+            self.assertFalse(form_field.label in str(res.content))
 
-    def test_highlights_correct_reaction(self):
+    def test_reaction_question_multi_submissions(self):
```
KaitCrawford, line 700: I have a number of concerns with the changes to this test. I'll schedule a call with you to discuss it rather than us going back and forth here.

test_settings.py
```diff
@@ -1,10 +1,10 @@
...
-        'ENGINE': 'django.db.backends.sqlite3',
-        'NAME': 'gem_test.db',
-    }
-}
+# DATABASES = {
+#     'default': {
+#         'ENGINE': 'django.db.backends.sqlite3',
+#         'NAME': 'gem_test.db',
+#     }
+# }
```
KaitCrawford, line 7: I don't think this should be commented out should it? I'm guessing you did it for your local testing?

moh-moola: @KaitCrawford Due to sqlite syntax issues when running migrations, I had to force test settings to use a postgres db

KaitCrawford: @moh-moola did you add other db settings to get the migrations running locally? They aren't in this PR

moh-moola: @KaitCrawford this is all that was needed

KaitCrawford: @moh-moola I'm worried this is going to break things for the rest of us locally.
Because without these settings then the app will still use sqlite because of these settings https://github.com/praekeltfoundation/molo-gem/blob/develop/gem/settings/base.py#L255
Did you maybe set the `DATABASE_URL` setting in your local environment to use postgres?

moh-moola: @KaitCrawford Nope, locally and on trav there are no test issues, I removed just the explicit declaration of the sqlite db conf

PR #1386, opened 2020-06-18
Form cleanup

Mitso:


Mitso:
@moh-moola Commenting out the reaction question slug, happy to run through how to best test that for this rq scenario.

PR #1394, opened 2020-07-22
update radio option unique ids

Mitso:


KaitCrawford:
@Mitso won't this give all radio button options for 1 question the same name? Is that ok?

Mitso:
@KaitCrawford No it wont. So the ids will now be a input-field question + option name 

e.g 
Question 1: 
yes = question_1_yes instead of id_yes
no = question_1_no instead of id_no

Question 2: 
yes = question_2_yes instead of id_yes
no = question_2_no instead of id_no

PR #1409, opened 2020-09-25
Set SESSION_COOKIE_SECURE setting to true, to prevent cookies being s…

moh-moola:
set SESSION_COOKIE_SECURE setting to true, to prevent cookies being s…ent over http


OMosimege (approved):
👍

PR #1410, opened 2020-11-24
Add PWA

codiebeulaine:


gem/templates/springster/base.html
```diff
@@ -94,7 +93,14 @@
     </noscript>
 </head>
   <body class="{% block body_class %}{% endblock %}" {% if request.LANGUAGE_CODE|language_bidi == True %}dir="rtl"{% endif %}>
-
+    <script>
+ if (!navigator.serviceWorker.controller) {
+     navigator.serviceWorker.register("/serviceworker.js").then(function(reg) {
+         console.log("Yo yo yo yo ");
+         console.log("Service worker has been registered: " + reg.scope);
```
KaitCrawford, line 99: :laughing: 
are you sure you want to leave these log lines in here?

gem/settings/base.py
```diff
@@ -574,8 +574,30 @@ def get_default_template(site_layout_base, site_layout):
     MEDIA_URL = "https://%s/" % AWS_S3_CUSTOM_DOMAIN
     DEFAULT_FILE_STORAGE = 'storages.backends.s3boto.S3BotoStorage'
 
-PWA_SERVICE_WORKER_PATH = join(
-    PROJECT_ROOT, 'templates', SITE_LAYOUT_BASE, 'serviceworker.js')
+PWA_APP_NAME = 'Springster-APP'
```
KaitCrawford, line 576: @codiebeulaine these settings are used for the culture brand sites as well right?
maybe we should make them configurable through the settings then?
although is that not possible because the manifest needs to be hard-coded?

KaitCrawford: Ah, I guess this is only loaded in the Springster templates anyway so maybe not neccessary.

KaitCrawford (approved):
:+1:

gem/settings/base.py
codiebeulaine: Yeah initially we only had PWA for springster. And also if you look below the other settings that were there are for sprignster

PR #1412, opened 2021-01-08
upgrade for django 3

codiebeulaine:


gem/static/manifest.json
```diff
@@ -1,15 +1,17 @@
...
+    "background_color": "#FFF",
+    "theme_color": "#493174",
+    "description": "Test app for Django and PWA",
+    "dir": "ltr",
+    "lang": "en-US",
+    "orientation": "portrait-primary",
+    "icons": [
+      {"src": "/static/img/appicons/springster_icon_96.png", "sizes": "96x96", "type": "image/png"},
+      {"src": "/static/img/appicons/springster_icon_144.png", "sizes": "144x144", "type": "image/png"},
+      {"src": "/static/img/appicons/springster_icon_192.png", "sizes": "192x192", "type": "image/png"}]
```
KaitCrawford: Why are you changing all this PWA stuff back?

gem/tests/test_context_processors.py
```diff
@@ -95,42 +75,6 @@ def test_returns_settings(self):
                 'REGISTRATION_URL': u'/profiles/register/',
                 'LOGOUT_URL': reverse('molo.profiles:auth_logout'),
                 'ENV': 'test_env',
-                'STATIC_URL': 'test_static_url',
+                'STATIC_URL': '/test_static_url',
```
KaitCrawford, line 78: is this a new django thing? that the static_url gets a '/' in front? because otherwise I don't see why this is being added, since we override this setting in a decorator and the context processor is just returning the setting.

gem/settings/base.py
```diff
@@ -229,7 +199,7 @@ def get_default_template(site_layout_base, site_layout):
 # We have multiple layouts: use `base`, `malawi` , `springster`, `rwanda`
 # 'chhaajaa' or 'yegna'
 # Change SITE_LAYOUT_BASE to switch between them.
-SITE_LAYOUT_BASE = environ.get('SITE_LAYOUT_BASE', 'base')
+SITE_LAYOUT_BASE = environ.get('SITE_LAYOUT_BASE', 'springster')
```
KaitCrawford: Have you decided to default to the springster templates or is a change you made for testing that needs to be reversed?

gem/views.py
```diff
@@ -53,75 +45,6 @@ def report_response(request, comment_pk):
...
-        site = request.site
-        language = getattr(request, 'LANGUAGE_CODE', settings.LANGUAGE_CODE)
-        if not hasattr(site, "oidcsettings"):
-            raise RuntimeError(
-                "Site {} has no settings configured.".format(site))
-        params.update({'theme': settings.THEME, 'language': language})
-        return params
-
-
 class RedirectWithQueryStringView(RedirectView):
```
KaitCrawford: is this view still imported somewhere? I see it was removed from the urls.py file.

gem/tests/test_admin.py
```diff
@@ -58,7 +58,7 @@ def test_comment_report_reasons_displaying_correctly(self):
 
     def test_admin_api(self):
         response = self.client.get(
-            '/admin/api/v2beta/pages/?child_of=1&for_explorer=1')
+            '/admin/api/v2beta/pages/')
```
KaitCrawford, line 61: Why are these query parameters not required anymore?

KaitCrawford (commented):
I'm concerned about the changes to serviceworker.js as well. Is that reverting some of the PWA work as well?

gem/tests/test_context_processors.py
codiebeulaine: I assumed so it was failing without it in the tests at least there was no / being added

gem/views.py
codiebeulaine: I don't think so, removing it

gem/settings/base.py
```diff
@@ -78,11 +56,11 @@
 
 MAINTENANCE_MODE = environ.get('MAINTENANCE_MODE', '') == 'true'
 ALLOWED_HOSTS = environ.get('ALLOWED_HOSTS', '*').split(",")
-INTERNAL_IPS = [
-    'localhost',
-    '.localhost',
-    '127.0.0.1'
-]
+INTERNAL_IPS = ("*", "codie")
```
KaitCrawford: is this right?

gem/settings/base.py
```diff
@@ -78,11 +56,11 @@
...
+RENDER_PANELS = True
+
+def return_true(soemthing):
+    return True
```
KaitCrawford: What is this little function for?

gem/settings/production.py
```diff
@@ -3,7 +3,7 @@
 
 # Disable debug mode
 
-DEBUG = False
+DEBUG = True
```
KaitCrawford: I don't think the prod settings should have DEBUG=True

KaitCrawford (commented):
Small queries on settings things

KaitCrawford (approved):
:+1: :cake: :chocolate_bar: :wine_glass: :beer: 
when the tests pass :)

PR #1416, opened 2021-05-19
Add shortlist winner

codiebeulaine:


KaitCrawford (approved):
:+1: In future we can probably just remove the chhaajaa templates instead of updating them.

PR #1422, opened 2022-01-31
add download scripts

codiebeulaine:


gem/management/commands/download_articles.py
```diff
@@ -0,0 +1,40 @@
...
+            article_data["subtitle"] = article.subtitle
+            if article.image:
+                article_data["image_name"] = article.image.title
+            body_string = ''
+            for block in article.body:
+                rendered = block.render_as_block()
+                if block.block_type == 'heading':
+                    rendered = "<h2>" + block.render_as_block() + "</h2>"
+                if block.block_type == 'numbered_list':
+                    rendered = block.render_as_block().replace("ul", "ol")
```
KaitCrawford, line 32: Why are we replacing `ul` with `ol` here?
Wouldn't that make unordered lists into ordered lists?
I think we might need the tags here as well. This will also change copy like `pull` to `poll` ect

gem/management/commands/download_articles.py
```diff
@@ -0,0 +1,40 @@
...
+            article_data["translation_pks"] = [page.pk for page in article.translated_pages.all()]
+            article_data["subtitle"] = article.subtitle
+            if article.image:
+                article_data["image_name"] = article.image.title
+            body_string = ''
+            for block in article.body:
+                rendered = block.render_as_block()
+                if block.block_type == 'heading':
+                    rendered = "<h2>" + block.render_as_block() + "</h2>"
+                if block.block_type == 'numbered_list':
```
KaitCrawford, line 31: is numbered_list the only list option here? is there no bulleted list option that we need to account for?

gem/management/commands/download_articles.py
```diff
@@ -0,0 +1,40 @@
...
+from django.core.management.base import BaseCommand
+from molo.core.models import ArticlePage
+from wagtail.admin.rich_text.converters.editor_html import EditorHTMLConverter
+
+
+class Command(BaseCommand):
+
+    def handle(self, *args, **options):
+        data = {}
+        articles = ArticlePage.objects.all()
```
KaitCrawford: Maybe we should do some ordering here? so that when we pull them into the new site they're more likely to be in the current ordering

gem/management/commands/download_locales.py
```diff
@@ -0,0 +1,19 @@
...
+from django.core.management.base import BaseCommand
+from molo.core.models import SiteLanguage
+
+
+class Command(BaseCommand):
+
+    def handle(self, *args, **options):
+        data = {}
+        langs = SiteLanguage.objects.all()
+        print(langs)
```
KaitCrawford: Stray print statement. Maybe you could change this into a log?

gem/management/commands/download_articles.py
```diff
@@ -0,0 +1,40 @@
...
+            article_data = {}
+            article_data["title"] = article.title
+            article_data["slug"] = article.slug
+            article_data["locale"] = article.language.locale
+            article_data["live"] = article.live
+            article_data["uuid"] = article.uuid
+            article_data["translation_pks"] = [page.pk for page in article.translated_pages.all()]
+            article_data["subtitle"] = article.subtitle
+            if article.image:
+                article_data["image_name"] = article.image.title
```
KaitCrawford, line 24: Should we be getting the image path as well?

gem/management/commands/download_articles.py
```diff
@@ -0,0 +1,40 @@
...
+                if block.block_type == 'heading':
+                    rendered = "<h2>" + block.render_as_block() + "</h2>"
+                if block.block_type == 'numbered_list':
+                    rendered = block.render_as_block().replace("ul", "ol")
+                body_string += (rendered + "\n")
+            article_data["body"] = body_string
+            article_data["featured_in_homepage"] = article.featured_in_homepage
+            if article.get_parent_section():
+                article_data["parent_section_slug"] = article.get_parent_section(locale=article_data["locale"]).specific.slug
+                article_data["parent_section_locale"] = article.get_parent_section(locale=article_data["locale"]).specific.language.locale
```
KaitCrawford, line 38: Should we get the pk as well?

gem/management/commands/download_articles.py
codiebeulaine: because it is a numbered list, not bullet point. it should be ordered

gem/management/commands/download_articles.py
codiebeulaine: no the bullet point lists don't need to be ordered so no need to change ul to ol

gem/management/commands/download_articles.py
codiebeulaine: yeah getting that

gem/management/commands/download_articles.py
codiebeulaine: we can but there is no use for it because the new pk will be different

gem/management/commands/download_articles.py
KaitCrawford: :facepalm: right :sweat_smile:

gem/management/commands/download_articles.py
KaitCrawford: oki doki

gem/management/commands/download_sections.py
```diff
@@ -0,0 +1,28 @@
...
+            section_data = {}
+            section_data["title"] = section.title
+            section_data["slug"] = section.slug
+            section_data["locale"] = section.language.locale
+            section_data["live"] = section.live
+            section_data["uuid"] = section.uuid
+            section_data["translation_pks"] = [page.pk for page in section.translated_pages.all()]
+            section_data["description"] = section.description
+            if section.image:
+                section_data["image_name"] = section.image.title
```
KaitCrawford, line 22: I think you need the image url here as well

KaitCrawford (commented):
What happens if there's an image block in the body of the article? will this still add the image url in a way that we can use it?

gem/management/commands/download_sections.py
codiebeulaine: if there is an image in the richtext or image block in the body, it gets added automatically, with all it's details

KaitCrawford: I think you were responding to my other comment? this one is about copying the url for the image used on the Section page

KaitCrawford (approved):
:+1:

PR #1423, opened 2022-02-09
remove pwa service worker

codiebeulaine:


KaitCrawford (approved):
:+1: when workflows are done

PR #1424, opened 2022-02-10
remove pwa

codiebeulaine:


KaitCrawford (approved):
:+1: 
This won't affect the Messenger plugin on ZA right? (just concerned about the FCM settings being removed)

